PR #1, opened 2026-04-21
docs(arch): v0 PRD and architecture — crew, runners, missions, signals, messages

yicheng47:
## Summary

Drafts the v0 PRD and architecture docs for Runners at `docs/arch/v0-prd.md` and `docs/arch/v0-arch.md`. Establishes the domain model, coordination design, and UI scope for the first shippable version.

### Domain model

- **Configuration layer** (persistent): **Crew**, **Runner**, **Orchestrator Policy**.
- **Runtime layer** (mission-scoped): **Mission** as the container; **Sessions** (one PTY per runner per mission) as first-class members; the coordination bus; orchestrator in-memory state; shared context.
- A Mission is to a Crew what a Session is to a Runner: the *run* of a *configuration*.

### Coordination primitives

- **Signal** (v0) — typed, orchestrator-routable notifications. Verb grammar (`review_requested`, `approved`, `blocked`).
- **Message** (v0) — prose, broadcast (`runners msg post`) or directed (`runners msg post --to <runner>`).
- **Inbox** (v0) — projection over message events: `to=null OR to=my_name`. `runners msg read` returns the caller's inbox.
- **Thread** (v0.x), **Fact** (v0.x), **Mention** / **Reaction** (v1) — documented with milestone tags.

### Delivery model

Signals are orchestrator-routed and drive `inject_stdin`, `ask_human`, etc. Messages are pull-based — runners read them on convention (system-prompt habit) and on signal-driven wake-ups. When `inject_stdin` fires, the orchestrator automatically appends the recipient's unread inbox summary so an urgent signal carries the async conversation along with it. No auto-nudge on direct messages.

### Session ergonomics

- PTY-backed via `portable-pty`; xterm.js in the webview for full TUI fidelity.
- **Human takeover is first-class**: the xterm pane is a real terminal; keystrokes pass through untouched.
- **Sessions outlive the UI**: closing the mission window does not kill sessions — agents keep running, the orchestrator keeps routing.
- PTY master writer serialized behind a mutex so human and orchestrator writes don't interleave.

### Event transport

One append-only NDJSON file per mission at `$APPDATA/runners/crews/{crew_id}/missions/{mission_id}/events.ndjson`. Per-mission scoping solves log rotation, crash-replay, and deletion. Signals and messages are both persisted as events with a `kind` discriminator.

### How runners emit events

Three layers, answering "how does the agent know to append to the event log":
1. System prompt (composed at spawn) describes the `runners` CLI and allowed signal types.
2. CLI exists on PATH with crew/mission/runner identity in env vars.
3. Role briefs reinforce usage with concrete examples.

### CLI surface (v0)

```
runners signal <type> [--payload <json>] [--correlation-id <id>] [--causation-id <id>]
runners msg    post <text> [--to <runner>] [--correlation-id <id>] [--causation-id <id>]
runners msg    read [--since <ts>] [--from <runner>]
```

### UI (text-only spec)

Mission control surfaces: runner list, focused terminal, signals pane, HITL panel, messages pane. The human sees every message in the mission (omniscient); runners see only their inbox.

## Commits in this PR

- scaffold: Session as first-class runtime member of Mission
- coordination: split events into signals + messages; defer threads/facts
- human takeover: elevate to first-class; sessions outlive the UI
- direct messaging: `--to` + inbox concept
- inbox delivery: pull-based; auto-nudge replaced with enriched stdin injection
- UI: text-only spec for the messages pane with All / Inbox modes

## Test plan

- [ ] Read `docs/arch/v0-prd.md` end-to-end and confirm the user journey (§5) is coherent and achievable.
- [ ] Read `docs/arch/v0-arch.md` end-to-end and confirm the concept model (§2) aligns with the PRD's vocabulary.
- [ ] Verify that every "v0" feature in the PRD has a corresponding mechanism in the arch doc.
- [ ] Flag any v0 scope creep before implementation starts.

### DIFF
--- a/docs/arch/v0-arch.md
+++ b/docs/arch/v0-arch.md
@@ -200,13 +200,18 @@ Every runner has an **inbox**: the subset of the mission's messages that are rel
 inbox(X) = all message events in the mission where to = null OR to = X
 ```
 
-`runners msg read` returns the calling runner's inbox, sorted by ULID (chronological).
+`runners msg read` returns the calling runner's inbox, sorted by ULID (chronological). `--since <ts>` restricts to messages newer than a given ULID/timestamp so agents can poll without re-reading history.
 
-This design keeps the storage model simple (one event log per mission, same as before) while giving each runner a clean "what's for me" view. Broadcasts end up in everyone's inbox; direct messages end up in exactly one.
+This design keeps the storage model simple (one event log per mission, same as before) while giving each runner a clean "what for me" view. Broadcasts end up in everyone's inbox; direct messages end up in exactly one.
 
-**Why LLM agents won't see direct messages unless they're told:** agents act on their prompt. They don't spontaneously poll the filesystem. A message landing in the inbox doesn't itself make the agent aware of it. We solve this with an orchestrator action: when a directed message arrives, the orchestrator nudges the recipient's stdin with a one-line hint ("new message from `coder` — run `msg read`"). The agent then reads its inbox as a normal tool call. See §5.5 action `nudge_recipient`.
+**The inbox is pull-based.** Messages are read when the recipient runs `msg read`; the system does not automatically interrupt a busy runner every time mail arrives. This is deliberate — not every direct message is urgent, and auto-interrupting on every DM would blur the signal/message split (§2.7.1 vs §2.7.2) and corrupt in-flight tool calls.
 
-The inbox is not a queue in the delete-on-read sense — messages stay in the log forever (well, for the mission). The "read" in `msg read` is lookup, not consumption. `msg read --since <ts>` lets agents fetch just what's new.
+The recipient learns to read its inbox through two mechanisms:
+
+1. **Convention** — the composed system prompt (§4.3) instructs every runner to check its inbox at natural task boundaries.
+2. **Signals as the urgent wake-up** — if a sender needs the recipient to drop everything, they emit a signal in addition to (or instead of) the message. The signal goes through the orchestrator's policy, which typically injects stdin. On any stdin injection, the orchestrator automatically appends a summary of the recipient's unread inbox messages (§5.5), so urgent wake-ups carry relevant conversation context with them.
+
+The inbox is not a queue in the delete-on-read sense — messages stay in the log forever (well, for the mission). The "read" in `msg read` is lookup, not consumption.
 
 #### 2.7.4 Thread *(v0.x)* — *scoped conversation*
 
@@ -300,7 +305,7 @@ MissionManager builds each runner's prompt from four parts:
 1. **The user-authored brief** (`runners.system_prompt`).
 2. **The mission brief** (`missions.goal_override` or `crews.goal`).
 3. **The roster** — crewmates' names, roles, one-line brief summaries.
-4. **Coordination notes** — how to use `runners signal` and `runners msg`, and the crew's allowed signal types.
+4. **Coordination notes** — how to use `runners signal` and `runners msg`, the crew's allowed signal types, and conventions for inbox checking.
 
 Example for a Reviewer:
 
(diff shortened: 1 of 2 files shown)


yicheng47 (commented):
Arch-only review, treating `docs/arch/v0-arch.md` as the source of truth:

**Findings**

- P1 `docs/arch/v0-arch.md:489`: enriched stdin depends on `last_read_ulid`, but `msg read` does not produce any defined event/receipt. "Infer from `--since`" is not reliable because runners can omit it or choose arbitrary values. Define an `inbox_read` audit event/high-watermark, or make injection summarize "recent inbox messages" instead of "unread."

- P1 `docs/arch/v0-arch.md:476`: `ask_human` emits `human_response`, but the arch never defines where the selected choice is stored or how policy matches it. Add a payload shape like `{ choice, question_id }` and explicit matching semantics for follow-up rules.

- P2 `docs/arch/v0-arch.md:230`: event naming is inconsistent. This says signals/messages become `signal_emitted` / `message_posted` events, while the canonical schema later uses `kind: "signal" | "message"`. Pick one representation.

- P2 `docs/arch/v0-arch.md:180`: direct addressing is based on runner names, but the data model only has display `name`, and examples mix `Coder` with `coder`. Add an immutable runner handle/slug or use `runner_id` for `from`/`to`, with display names kept separate.

- P2 `docs/arch/v0-arch.md:393`: the append atomicity claim relies on `PIPE_BUF`, which applies to pipes, not regular files. If concurrent CLI appends are core to correctness, specify a single `write(2)` call plus local filesystem assumptions, or use file locking/a single writer.

No arch-only formatting or whitespace issues found.

yicheng47:
Pushed `013d233` addressing all five findings.

**P1 #1 — watermark is now authoritative, not inferred.**
Every `runners msg read` invocation ends by emitting an `inbox_read` signal with `payload.up_to = <max ULID returned>`; the orchestrator advances the per-runner watermark from that event. `stdin_injected` events also carry `payload.watermark`, so an enriched injection counts as "shown" and subsequent signals won't re-summarize the same messages. The watermark is rebuilt from the log on crash-replay — never inferred from `--since`. See §5.5.1.

**P1 #2 — `ask_human` payload shapes + matching semantics specified.**
New §5.5.0 defines both `human_question` and `human_response` with full example JSON. `human_response.payload = { question_id, choice }`; `correlation_id` = the triggering signal's id; `causation_id` = the `human_question.id`. Two matching strategies documented: by `choice` value (v0 default), and by `correlated_with` (deferred to v0.x for the concurrent-question case).

**P2 #1 — single discriminator via `kind`.**
§2.8 rewritten: there is no `signal_emitted` or `message_posted`; `kind` is authoritative. For `kind: "signal"` the `type` field is the verb (`review_requested`, `changes_requested`, ...); for `kind: "message"` `type` is omitted and prose lives in `payload.text`.

**P2 #2 — runner identity split into `handle` and `display_name`.**
`handle` is a lowercase slug, immutable, unique within crew — used everywhere addressing happens (`from`/`to`, `--to`, policy rules). `display_name` is a free-form UI label. SQL updated with `UNIQUE (crew_id, handle)`. Env var renamed `RUNNERS_RUNNER_NAME` → `RUNNERS_RUNNER_HANDLE`. The composed-prompt example (§4.3) now shows both forms with an explainer; the PRD user journey introduces runners as `coder` (Coder) / `reviewer` (Reviewer). CLI validates `--to <handle>` against the crew's handles.

**P2 #3 — atomicity: dropped the `PIPE_BUF` claim; specified flock + single-write.**
New §5.1.1 "Concurrent-write correctness": `O_APPEND | O_WRONLY | O_CREAT` + `flock(LOCK_EX)` + single `write(2)` of the JSON line + `\n` + `close`. Explicitly notes `PIPE_BUF` applies to pipes not regular files, and that small-write atomicity on regular files is filesystem-specific — hence the lock. Documents filesystem requirements (local POSIX; NFS/SMB/iCloud unsupported) with a startup check.

All changes are in a single commit so the delta is easy to review.

PR #3, opened 2026-04-21
docs(impls): v0 MVP implementation plan

yicheng47:
## Summary
- Adds \`docs/impls/v0-mvp.md\` — umbrella plan for the v0 MVP.
- Breaks the first end-to-end vertical slice into **11 chunks** (C1–C11) with a clear dependency graph; each chunk is its own squash-merged PR into a shared \`feature/v0-mvp\` branch.
- Documents the demo "definition of done" (lead dispatches → worker replies → ask_human roundtrip → reopen + replay), explicit out-of-scope list, and commit / branching conventions.

## Test plan
- [ ] Skim \`docs/impls/v0-mvp.md\` and confirm the chunking feels right before C1 starts
- [ ] Sanity-check the out-of-scope list against \`v0-prd.md\`

### DIFF
--- a/docs/arch/v0-arch.md
+++ b/docs/arch/v0-arch.md
@@ -103,6 +103,8 @@ The persistent "who's on the team and how they work together" record. A crew has
 
 **Every crew must have exactly one lead runner.** The lead is the human's counterpart in the crew: mission goals and broadcast human messages route to the lead by default, and the lead dispatches work to the other runners via directed messages. This is a hard invariant — a crew with zero runners or zero leads is invalid and cannot start a mission. The first runner added to a new crew becomes lead automatically; the user can reassign lead between existing runners, but cannot remove the lead runner without first designating a replacement. The lead is a routing convention, not a privileged capability: any runner can emit signals, post directed messages, and trigger orchestrator actions. Lead only governs *where inbound-from-human traffic lands by default*.
 
+**Lead is also the default HITL gateway.** When a worker needs human input, it does not ask the human directly — it sends a directed message to the lead, who decides whether to answer from their own context or escalate to the human via `ask_human`. If the lead escalates, the human's answer flows back through the lead, who forwards it to the original worker as another directed message. This keeps the human's attention focused on one interlocutor and lets the lead absorb, filter, or batch worker questions. See §5.5.0 for the protocol details (the `on_behalf_of` payload field and the forwarding flow). Workers *can* emit `ask_human` directly as a fallback — it's not forbidden at the protocol layer — but the default runner system prompt instructs them to route through the lead.
+
 Lifecycle: created by the user, edited freely, deleted when no longer needed. Persisted in SQLite.
 
 ### 2.3 Runner — *one configured agent*
@@ -533,7 +535,8 @@ Every action emits at least one audit signal. Each audit payload carries `trigge
     "question_id": "01HG...",                  // = this event's id; echoed here for convenience
     "triggered_by": <triggering-signal.id>,    // e.g. the changes_requested signal's id
     "prompt": "Reviewer requested changes. Accept or override?",
-    "choices": ["accept", "override"]
+    "choices": ["accept", "override"],
+    "on_behalf_of": "@impl"                    // optional; see "Lead-mediated asks" below
   }
 }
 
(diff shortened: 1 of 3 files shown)


yicheng47:
Review findings:

1. Medium: `docs/impls/v0-mvp.md:204` makes directed messages trigger `inject_stdin`, but the architecture says messages are pull-based and do not trigger orchestrator actions in v0. Implementing this plan would turn every DM into an interruption, bypassing the signal-as-wake-up design and risking PTY/tool-call corruption. The rule should be signal-driven, with unread messages included only when a signal causes injection.

2. Medium: `docs/impls/v0-mvp.md:68` defines a C1 schema that does not support later chunks or the arch/PRD model. It omits `crews.goal`, `orchestrator_policy`, `signal_types`, runner `runtime/command/working_dir/env`, and the `sessions` table, while later chunks depend on policy loading, signal validation, session lifecycle, and replay. Following C1 literally will force avoidable migrations/refactors by C6-C9.

3. Medium: `docs/impls/v0-mvp.md:11` puts “runner templates” in the MVP demo and UI chunks, but the PRD marks runner templates/presets as out of scope and defines runner CRUD as crew-scoped. Either the plan needs to rename these to crew-scoped runners/slots, or it needs to add a real template data model and update the PRD.

4. Low: `docs/impls/v0-mvp.md:74` describes `UNIQUE(crew_id) WHERE lead = 1` inside the table definition. SQLite partial uniqueness must be a separate partial index, e.g. `CREATE UNIQUE INDEX ... ON runners(crew_id) WHERE lead = 1`.

Validation: refreshed `main` and `docs/v0-mvp-plan`; `git diff --check origin/main...HEAD` is clean. No runtime tests run because this PR is documentation-only.

yicheng47:
Follow-up review findings after the update:

Note: the PRD now appears stale. For the follow-up work, I would treat `docs/arch/v0-arch.md` as the canonical source of truth and either update the PRD to match it or mark the stale PRD sections as superseded.

1. Medium: `docs/impls/v0-mvp.md:210` introduces `signal human_direct (to: <handle>)`, but the event schema still says signals always have `to = null` in v0 and only messages use directed `to` (`docs/arch/v0-arch.md:442`). Either target should move into `payload`, or the schema needs to explicitly allow targeted human-originated signals.

2. Medium: `docs/impls/v0-mvp.md:211` depends on runner-emitted built-in signals like `ask_lead` / `ask_human`, but the CLI validates signal types against `signal_types.json` (`docs/arch/v0-arch.md:478`). C1/C2 don’t say to seed built-in signal types into every crew’s allowlist, so the demo path can fail with the CLI rejecting `ask_lead`.

3. Medium: `docs/impls/v0-mvp.md:70` still defines a schema that diverges from the architecture data model: `missions(title, goal, cwd)` vs arch `goal_override`, and runner `args/cwd/env` vs arch `args_json/working_dir/env_json`. The update fixed missing coverage, but implementers still have two incompatible schema sources. Since the arch is canonical, the implementation plan should align with its field names or explicitly document intentional deviations.

4. Medium: The lead-mediated HITL flow updates arch/impl, but the PRD still describes the old direct `changes_requested -> ask_human -> inject Coder` flow and still includes `correlation_id` / `causation_id` CLI fields (`docs/arch/v0-prd.md:76`, `docs/arch/v0-prd.md:154`). If the PRD is intentionally stale, that should be called out; otherwise it needs to be updated to match the architecture.

Validation: refreshed PR head; `git diff --check origin/main...origin/docs/v0-mvp-plan` is clean.

yicheng47:
Follow-up review after the latest update:

The last round’s main issues are addressed: `human_said` now uses `payload.target`, built-in signal types are seeded, C5 exports the sidecar before sessions spawn, schema names are closer to the arch, and the PRD now explicitly says the arch doc is canonical.

One remaining finding:

1. Medium: `docs/impls/v0-mvp.md:69` says C1 field names match the arch data model exactly, but the schema still differs from canonical arch §7.1: it adds `crews.purpose`, `missions.title/cwd`, `sessions.handle/pid`, and runner `lead/position`. Some of those may be needed, but with the new “arch wins” rule in the PRD, either arch §7.1 should be updated to include them or the implementation plan should explicitly call them intentional implementation additions.

Minor stale wording remains at `docs/impls/v0-mvp.md:94`: “runner templates” should be “crew-scoped runners.”

Validation: refreshed PR head; `git diff --check origin/main...origin/docs/v0-mvp-plan` is clean.

PR #4, opened 2026-04-22
feat(db): C1 — schema + shared types

yicheng47:
## Summary

Chunk 1 of the v0 MVP umbrella — lays down the SQLite schema and the Rust/TS type surface every later chunk consumes. See `docs/impls/v0-mvp.md` §C1.

- **Schema** (`src-tauri/migrations/0001_init.sql`): `crews`, `runners`, `missions`, `sessions` — arch §7.1 verbatim — plus the `one_lead_per_crew` partial unique index (arch §2.2 invariant).
- **Signal allowlist seeding**: `crews.signal_types` has a SQL `DEFAULT` carrying the seven built-in MVP signal types (`mission_goal`, `human_said`, `ask_lead`, `ask_human`, `human_question`, `human_response`, `inbox_read`). Without this, the CLI's arch §5.3 Layer 2 validation will reject built-ins at spawn time. `DEFAULT_SIGNAL_TYPES` and `default_signal_types_json()` also exposed from `db.rs` for C5's sidecar export.
- **DB layer** (`src-tauri/src/db.rs`): r2d2 connection pool, WAL + `foreign_keys = ON` + `busy_timeout = 5000` on every connection, hand-rolled migrations runner (tracked in `_migrations`), bootstrapped in `lib.rs::setup`.
- **Rust types** (`src-tauri/src/model.rs`): `Crew`, `Runner`, `Mission`, `Session`, `Event`, `EventKind`, `SignalType`, `MissionStatus`, `SessionStatus`. Event envelope follows arch §5.2 — `type` field is `skip_serializing_if` None so messages omit it while signals include it.
- **TS types** (`src/lib/types.ts`): hand-synced mirror. No `ts-rs` — MVP-light.

## Test plan

`cargo test --lib` — 10/10 passing:

- [x] migrations bootstrap all four tables
- [x] new crew seeded with default signal allowlist
- [x] `one_lead_per_crew` rejects a second lead in the same crew
- [x] `one_lead_per_crew` permits leads across different crews
- [x] `UNIQUE(crew_id, handle)` enforced
- [x] JSON blob columns roundtrip (`orchestrator_policy`, `signal_types`, `args_json`, `env_json`)
- [x] cascade delete removes dependent runner rows
- [x] reopening the DB re-runs migrations idempotently (each applied exactly once)
- [x] signal event envelope roundtrips with `type` preserved and `to: null`
- [x] message event omits `type` when serialized

Other gates:
- [x] `cargo check` — clean, zero warnings
- [x] `pnpm exec tsc --noEmit` — clean

## Out of scope (later chunks)

- No Tauri commands yet — C2 adds `crew_*` / `runner_*` with the lead invariant enforced in Rust as well as the DB.
- No event log writer — C4 ships `EventLog::{open, append, read_from}` + the ULID generator.
- No PTY, orchestrator, or UI.

Part of the v0 MVP umbrella → merges into `feature/v0-mvp`.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,0 +1,2614 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+
+  .:
+    dependencies:
+      '@tailwindcss/vite':
+        specifier: ^4.2.1
+        version: 4.2.4(vite@7.3.2(jiti@2.6.1)(lightningcss@1.32.0))
+      '@tauri-apps/api':
+        specifier: ^2
+        version: 2.10.1
+      '@tauri-apps/plugin-dialog':
+        specifier: ^2.6.0
+        version: 2.7.0
+      '@tauri-apps/plugin-fs':
+        specifier: ^2.4.5
+        version: 2.5.0
+      '@tauri-apps/plugin-opener':
+        specifier: ^2
+        version: 2.5.3
+      '@tauri-apps/plugin-process':
+        specifier: ^2.3.1
+        version: 2.3.1
+      '@tauri-apps/plugin-shell':
+        specifier: ^2
+        version: 2.3.5
+      lucide-react:
+        specifier: ^0.577.0
+        version: 0.577.0(react@19.2.5)
+      react:
+        specifier: ^19.1.0
+        version: 19.2.5
+      react-dom:
+        specifier: ^19.1.0
+        version: 19.2.5(react@19.2.5)
+      react-router-dom:
+        specifier: ^7.13.1
+        version: 7.14.2(react-dom@19.2.5(react@19.2.5))(react@19.2.5)
+      tailwindcss:
+        specifier: ^4.2.1
+        version: 4.2.4
+    devDependencies:
+      '@eslint/js':
+        specifier: ^9.39.4
+        version: 9.39.4
+      '@tauri-apps/cli':
+        specifier: ^2
+        version: 2.10.1
+      '@types/react':
+        specifier: ^19.1.8
+        version: 19.2.14
+      '@types/react-dom':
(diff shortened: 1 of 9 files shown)


yicheng47 (commented):
**Findings**

- **Medium** `src-tauri/src/db.rs:76`: migrations are applied outside a transaction, then recorded afterward. If the app crashes or errors after `execute_batch(sql)` but before inserting into `_migrations`, the DB can be left with tables created but version `0`; next startup reruns `0001_init.sql` and fails on `CREATE TABLE crews`. Wrap each migration plus the `_migrations` insert in one transaction, ideally `BEGIN IMMEDIATE`.
- **Low** `src-tauri/src/db.rs:169`: `cargo fmt --check` fails on formatting in `db.rs`.

**Checks**

- Passed: `cargo test --lib`, `cargo check`, `pnpm exec tsc --noEmit`, `pnpm run lint`
- Failed: `cargo fmt --check`
- GitHub: `gh pr checks 4` reports no checks configured.

yicheng47:
Addressed both findings + bundled the lockfile:

- **Medium (db.rs migrations atomicity)** — each migration + its `_migrations` insert now runs in a single `BEGIN IMMEDIATE` transaction (`transaction_with_behavior(TransactionBehavior::Immediate)` + `tx.commit()`). A crash after DDL but before the bookkeeping row now rolls back the schema change, so the next startup retries the same version instead of replaying onto a partially-migrated DB.
- **Low (cargo fmt)** — ran `cargo fmt`; `cargo fmt --check` now clean.
- **pnpm-lock.yaml** — added. Not gitignored, and `Cargo.lock` is already tracked; Tauri repo is a deployable app (not a library), so pinning transitive JS versions matches the existing convention.

Gates rerun: `cargo fmt --check`, `cargo check`, `cargo test --lib` (10/10), `pnpm exec tsc --noEmit` — all pass.

PR #9, opened 2026-04-22
feat(ui): crew + runner config surface (C3)

yicheng47:
## Summary

C3 of the v0 MVP umbrella — wires the C2 Tauri commands to a usable config UI.

- **Crews page** — list/create/delete crews. Grid of cards with runner counts.
- **Crew detail** — ordered runner rows, LEAD badge, Set-as-lead, Edit, Delete, HTML5 drag-reorder. Optimistic reorder, reconciles with \`runner_reorder\` response.
- **Add runner modal** — handle / display_name / role / runtime / command / args / cwd / system_prompt. Client-side handle regex mirrors Rust \`validate_handle\`. Flags auto-LEAD when adding to an empty crew.
- **Runner edit drawer** — slide-over, handle shown read-only (immutable per arch §2.2, §5.2).
- **\`src/lib/api.ts\`** — typed \`invoke\` wrappers for all 12 C2 commands.
- **UI primitives** — Button / Field / Modal / Drawer in \`src/components/ui/\`.

Scope note: no standalone Runners page in MVP (runners are crew-scoped per PRD §3). No env editor yet — deferred to v0.x.

## Test plan

- [x] \`pnpm run lint\` clean
- [x] \`cargo check\` clean in \`src-tauri/\`
- [ ] \`pnpm run tauri dev\` — create a crew, add two runners, reassign lead, delete lead, confirm auto-promotion by position
- [ ] Drag-reorder two runners, confirm server-side positions persist
- [ ] Handle regex rejects \`BadHandle\`, \`has space\`, \`-lead\` inline; server still catches bypasses


yicheng47:
Review findings:

- P2 `src/pages/CrewEditor.tsx`: reorder requests are not serialized. A user can drop twice before the first `runner_reorder` returns; whichever request commits last wins, even if it represents an older UI order. Disable/queue drag reorder while one request is in flight, or serialize updates.
- P2 `src/pages/CrewEditor.tsx`: initial load errors are hidden as "Crew not found." Because the error banner only renders after `crew !== null`, any `crew_get` or `runner_list` failure during the first load loses the real backend error.
- P3 `src/pages/CrewEditor.tsx`, `src/pages/Crews.tsx`: hover-hidden actions stay focusable/clickable while invisible. Keyboard users tab into invisible buttons, and touch users can trigger undiscoverable delete/edit/lead actions. Add `group-focus-within` visibility and pointer-event handling, or keep actions visible.

Checks run:

- `pnpm lint`: passed
- `pnpm exec tsc --noEmit`: passed
- `cargo test`: passed
- `pnpm build`: blocked locally by Node `18.20.8` while Vite requires `20.19+` or `22.12+`, plus a missing Tailwind optional native binding.

yicheng47:
Addressed all three findings (e9088b5):

- **P2 reorder race** — `onCommitReorder` now early-returns if `reorderInFlight.current` is true. Row `draggable` flips to `false` while a request is pending (primary gate), and the re-entrancy check on the handler is the belt-and-suspenders backstop. Commented inline so future-me knows why both exist.
- **P2 initial-load errors** — added a `loaded` flag; if the first `refresh()` throws we render a red banner with the real error instead of falling through to "Crew not found." The latter now only fires when the server says the crew actually doesn't exist.
- **P3 focusable-while-invisible actions** — replaced `opacity-0 group-hover:opacity-100` with `opacity-60` baseline + full opacity on `group-hover` and `group-focus-within` (and `focus-visible` on the crew-card delete button). Controls stay visible-but-dim at rest, keyboard users can tab to them, touch users don't trigger phantoms.

Checks: `pnpm exec tsc --noEmit` ✓, `pnpm run lint` ✓.

yicheng47:
Amended (f220001):

- **AppShell + Sidebar** — persistent 240px sidebar (quill-style) with `runners` wordmark and Runners / Crews / Missions nav. Runners and Missions render disabled (scope-excluded in v0 MVP) so the shell keeps the design's shape. `data-tauri-drag-region` on top.
- **Home removed** — `/` now redirects to `/crews`; `src/pages/Home.tsx` deleted. The standalone "Open crews" landing was out of spec.
- **Crews page** rebuilt against frame `nqOot`: header (title + subtitle + `+ New crew`), single-column full-width card stack, trailing dashed `+ Add another crew` empty-state card. Creating a crew navigates straight into its editor.
- **Crew Detail** rebuilt against frame `CUKjM`: top toolbar with `← Crews` breadcrumb + inline editable name field + Save + Start mission (disabled, arrives in C11). Purpose block, then numbered Slots list with `⋮⋮` grip + text-link actions (Set as lead · Edit · Remove). Template palette omitted — no runner templates in MVP per PRD §3.
- **AddSlotModal** updated to frame `sYprG`: JetBrains Mono `@`-prefixed handle input, subtitle in header, `Add slot` label, consistent 24/20/16 padding rhythm.
- **Overlay** (Modal + Drawer) now uses `rounded-xl`, `bg-black/35` backdrop, and the design's `#FAFAFA` footer / `#E5E5E5` borders.

Also bundled in this push (from the earlier Node-version blocker):
- **`.nvmrc`** (`22`) unignored + committed; `package.json` `engines` pins Node ≥22.12 / pnpm ≥9.
- **CI** switched from `node-version: 20` to `node-version-file: .nvmrc` so GH Actions tracks the same pin.

Checks: `pnpm exec tsc --noEmit` ✓, `pnpm run lint` ✓, `cargo check` ✓.

yicheng47:
Follow-up review on `8962a12`:

- P2 `src/pages/Crews.tsx`: `CrewCard` is a clickable `div role="button"` that contains a real `Delete` button. Mouse delete works because the click stops propagation, but keyboard activation bubbles to the parent `onKeyDown`; pressing Enter/Space while focused on Delete opens the crew instead of reliably deleting it. Make the card/open affordance and delete button siblings, or ignore bubbled key events from child controls.

Checks run:

- `pnpm run lint`: passed, with Node engine warning
- `pnpm exec tsc --noEmit`: passed
- `pnpm run build`: passed, though Vite warns current Node is `18.20.8`
- `cargo test`: passed

PR #10, opened 2026-04-22
feat(event-log): NDJSON append + parse + monotonic ULID (C4)

yicheng47:
## Summary
- Adds `event_log::EventLog` (open / append / read_from) — append-only NDJSON log with exclusive `flock` per write so concurrent `runners signal` / `runners msg post` invocations interleave at whole-line granularity (arch §5.4).
- Adds `event_log::UlidGen` — mutex-wrapped monotonic ULID generator. Same-ms collisions resolve by incrementing the 80-bit random payload so ordering stays lexicographic (arch §5.2).
- Adds `event_log::path` helpers matching the `$APPDATA/runners/crews/{id}/missions/{id}/events.ndjson` tree in arch §7.2, plus the `signal_types.json` sidecar path C5 will export.
- No watcher — that's C7. This chunk only ships primitives.

## Test plan
- [x] `cargo test --lib` — 37 passing (8 new: ULID monotonicity + uniqueness under 8-thread fan-in, 2000-event append/readback with no interleave, empty-log read, embedded-newline refusal, resume-by-offset, filesystem layout).
- [x] `cargo check` clean.

Part of the v0 MVP umbrella (C4). See `docs/impls/v0-mvp.md`.


yicheng47:
Found three issues.

1. High: ULID ordering is not serialized with append order.
   `src-tauri/src/event_log/ulid.rs:17` is only an in-process mutex, while `src-tauri/src/event_log/log.rs:64` accepts a preassigned event ID and only locks around the write. Concurrent CLI processes can append a lower ULID after a higher ULID has already been read/watermarked, causing later `msg read --since <max_ulid>` flows to permanently skip that late event. ID assignment or ordering enforcement needs to happen under the same cross-process serialization point as the file append, or watermarks need to be offsets instead of ULIDs.

2. Medium: partial writes can poison the log permanently.
   `src-tauri/src/event_log/log.rs:81` uses `write_all`, which does not preserve the architecture’s “exactly one write(2)” invariant and does not roll back if bytes were partially written before an error. `src-tauri/src/event_log/log.rs:107` only skips a partial line while it is still the final EOF line; after the next append, that partial prefix plus the next JSON line becomes a malformed newline-terminated line and replay will fail at `serde_json::from_str`.

3. Medium: the C4 primitive is not actually reusable by the planned CLI crate.
   `src-tauri/src/lib.rs:5` declares `event_log` as a private Tauri module, but C9 calls for a separate `cli/` crate reusing the C4 event log code directly. This should probably be a shared crate, or at minimum a public API with public model/error types.

Validation: `cargo test --lib`, `cargo check`, and `git diff --check origin/feature/v0-mvp...HEAD` all pass locally.

yicheng47:
Follow-up review findings after the latest patch:

1. Medium: dangling partial tail can still poison the log after a crash.
   `crates/runners-core/src/event_log/log.rs:187` treats the final bytes as the last event even when they are not newline-terminated. If a process dies after writing JSON bytes but before the trailing `\n`, `EventLog::open` can fail on a truncated tail, or worse, accept a valid unterminated JSON event and the next `append` (`crates/runners-core/src/event_log/log.rs:132`) glues the next event onto it. The append path should find/truncate any dangling non-newline tail under the lock before scanning the last ID or writing.

2. Low: the local Rust test target no longer runs the new core tests.
   `Makefile:36` still runs `cd src-tauri && cargo test`, which only runs the 27 app tests and skips the 14 `runners-core` tests where the C4 behavior now lives. CI was updated to `cargo test --workspace`, but `make test-rust` / `make test` should match that.

Validation: `cargo test --workspace`, `cargo check --workspace`, `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `git diff --check origin/feature/v0-mvp...HEAD` all pass. I also confirmed `cd src-tauri && cargo test` skips `runners-core`.

PR #11, opened 2026-04-22
feat(mission): lifecycle commands + opening events (C5)

yicheng47:
## Summary
- Adds \`commands::mission\` with \`mission_start / mission_stop / mission_list / mission_get\`.
- \`mission_start\` is the config→runtime crystallization point: validates the crew (≥1 runner, exactly one lead per the arch §2.2 invariant), creates the mission row and on-disk mission directory, atomically exports the crew's \`signal_types\` allowlist to the sidecar the CLI reads (arch §5.3 Layer 2), then emits the two opening events to \`events.ndjson\`:
  - \`mission_start\` (\`from: "system"\`, payload \`{ title, cwd }\`)
  - \`mission_goal\` (\`from: "human"\`, payload \`{ text }\`) — orchestrator's built-in rule in C8 will route this to the lead via \`inject_stdin\`.
- \`mission_stop\` flips status to \`completed\`, sets \`stopped_at\`, appends \`mission_stopped\`.
- Threads \`app_data_dir\` into \`AppState\` so mission commands can resolve event-log paths via \`runners_core::event_log\`.
- No PTYs yet — that's C6.

## Test plan
- [x] \`cargo test --workspace\` — 50 passing (7 new: empty-crew / leadless-crew / empty-title rejection, two-opening-events + sidecar written, goal_override beats crew default, stop marks completed + appends event, already-stopped rejection, list filters by crew + orders desc).
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\` clean.
- [x] \`cargo fmt --all --check\` clean.

Part of the v0 MVP umbrella (C5). See \`docs/impls/v0-mvp.md\`.

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"580cc5f7-5532-46ef-8de6-f5ef4a683b26","pid":9431,"acquiredAt":1776865184309}
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,1 +1,0 @@
-{"sessionId":"580cc5f7-5532-46ef-8de6-f5ef4a683b26","pid":9431,"acquiredAt":1776865184309}
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ dist-ssr
 .claude/settings.local.json
 src-tauri/target
 /target
+.claude/scheduled_tasks.lock
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -28,6 +28,6 @@ r2d2_sqlite = "0.25"
 chrono = { workspace = true }
 thiserror = { workspace = true }
 ulid = { workspace = true }
-
-[dev-dependencies]
+# Regular dep (not dev-only) because the sidecar writer needs
+# NamedTempFile::persist for a Windows-safe atomic replace.
 tempfile = "3"
--- a/src-tauri/src/commands/mission.rs
+++ b/src-tauri/src/commands/mission.rs
@@ -1,0 +1,598 @@
+// Mission lifecycle — start, stop, list, get.
+//
+// A mission is the runtime container: it owns a directory, an NDJSON event
+// log, and a set of sessions (spawned in C6). This module only does the
+// bookkeeping layer — no PTYs yet.
+//
+// `mission_start` is the point where config crystallizes into runtime:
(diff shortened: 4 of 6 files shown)


yicheng47:
Review findings:

1. High: mission rows can be committed without the required event log state.
   In `src-tauri/src/commands/mission.rs:153`, `start` inserts the `running` mission before creating/exporting the sidecar and appending `mission_start` / `mission_goal`; if any filesystem or `EventLog::append` call fails, the command returns an error but leaves a visible running mission with no valid startup log. `stop` has the same shape at `src-tauri/src/commands/mission.rs:222`: it marks completed before appending `mission_stopped`, so a log failure leaves the row completed without the terminal event. The DB transition should be committed only after the required log work succeeds.

2. Medium: `mission_start` does not enforce one live mission per crew.
   The architecture states this v0 constraint in `docs/arch/v0-arch.md:146`, but `start` (`src-tauri/src/commands/mission.rs:123`) only validates runners/lead and then inserts a new `running` mission. A second invoke can create another live mission for the same crew, which C6/C7 assume is out of scope.

3. Medium: sidecar replacement is not portable to Windows.
   `write_signal_types_sidecar` (`src-tauri/src/commands/mission.rs:270`) uses `std::fs::rename(tmp, target)` over an existing `signal_types.json`. That replaces on Unix, but fails on Windows when the destination exists, so the second mission start for a crew can fail on the NSIS target.

Validation passed: `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`, and `git diff --check origin/feature/v0-mvp...HEAD`.

yicheng47:
Follow-up review finding after the latest patch:

1. Medium: concurrent `mission_stop` can append duplicate terminal events.
   `src-tauri/src/commands/mission.rs:238` reads and checks `Running` before opening the transaction, then updates by id only at `src-tauri/src/commands/mission.rs:253`. Two stop calls can both observe `running`; the first commits `mission_stopped`, then the second still updates the row and appends another `mission_stopped`. The status check/update should be atomic, e.g. `UPDATE ... WHERE id = ? AND status = 'running'` and require `affected == 1`.

The previous three findings are otherwise addressed: start/stop rollback behavior is improved, live mission creation is checked, and `tempfile::persist` does use replace semantics on Windows.

Validation passed: `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`, `cargo check --workspace`, and `git diff --check origin/feature/v0-mvp...HEAD`. I ignored the unrelated untracked `.claude/scheduled_tasks.lock`.

PR #12, opened 2026-04-22
feat(session): PTY session runtime + debug page (C6)

yicheng47:
## Summary
- \`session::SessionManager\` spawns a PTY child per runner via portable-pty. Env vars (RUNNERS_CREW_ID / MISSION_ID / RUNNER_HANDLE / EVENT_LOG + \$MISSION_CWD) are set at spawn time so C9's CLI can pick them up.
- One reader thread per session pumps bytes → \`SessionEvents::output\`, waits the child on EOF, updates the \`sessions\` row to stopped/crashed, and emits \`SessionEvents::exit\`.
- \`SessionEvents\` trait decouples the PTY layer from Tauri so tests run with a capture fake. Prod uses \`TauriSessionEvents(AppHandle)\` → \`session/output\` + \`session/exit\` Tauri events.
- \`mission_start\` now spawns one PTY per runner; \`mission_stop\` kills them before flipping the mission row.
- **Debug page at \`/debug\`** (scratch, to be removed when C10 lands) — picks a crew, starts a mission, renders one pane per runner with live output + stdin input + kill button. Exists so runtime reality surfaces before the polished workspace UI goes in.

## Test plan
- [x] \`cargo test --workspace\` — 57 passing (3 new PTY tests: end-to-end echo, cat + stdin roundtrip, unknown-session error).
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\` clean.
- [x] \`pnpm tsc --noEmit\` + \`pnpm lint\` clean.
- [ ] Manual: \`make dev\` → navigate to \`/debug\`, pick a crew with \`/bin/sh\` as a runner command, start mission, send \`echo hi\`, see output stream back, kill, see exit land. Then try a real agent (\`claude-code\`, \`codex\`) to derisk C10.

Part of the v0 MVP umbrella (C6). See \`docs/impls/v0-mvp.md\`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3280,6 +3280,7 @@ name = "runners"
 version = "0.1.0"
 dependencies = [
  "chrono",
+ "libc",
  "portable-pty",
  "r2d2",
  "r2d2_sqlite",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -829,6 +829,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "downcast-rs"
+version = "1.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75b325c5dbd37f80359721ad39aca5a29fb04c89279657cffdda8736d0c0b9d2"
+
+[[package]]
 name = "dpi"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -875,7 +881,7 @@ dependencies = [
  "rustc_version",
  "toml 0.9.12+spec-1.1.0",
  "vswhom",
- "winreg",
+ "winreg 0.55.0",
 ]
 
 [[package]]
@@ -1001,8 +1007,19 @@ version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "38e2275cc4e4fc009b0669731a1e5ab7ebf11f469eaede2bab9309a5b4d6057f"
 dependencies = [
- "memoffset",
+ "memoffset 0.9.1",
  "rustc_version",
+]
+
+[[package]]
+name = "filedescriptor"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e40758ed24c9b2eeb76c35fb0aebc66c626084edd827e07e1552279814c6682d"
+dependencies = [
+ "libc",
+ "thiserror 1.0.69",
+ "winapi",
 ]
 
 [[package]]
@@ -1842,6 +1859,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ioctl-rs"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f7970510895cee30b3e9128319f2cefd4bde883a39f38baa279567ba3a7eb97d"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "ipnet"
 version = "2.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


yicheng47:
Review findings:

1. High: `mission_start` can fail while leaving a running mission and already-spawned sessions behind. It commits the mission in `src-tauri/src/commands/mission.rs:328`, then loops runners with `spawn(...)?` in `mission.rs:345-352`. If runner 2 has a bad command after runner 1 spawned, the command returns an error but runner 1 keeps running and the mission row stays `running`, blocking future starts for that crew. The command should either make spawn all-or-nothing/rollback-kill on failure, or return a successful partial-start state with explicit recovery semantics.

2. High: `mission_stop` does not actually wait for sessions to die before completing the mission. `mission_stop` calls `kill_all_for_mission`, ignores its result, then immediately marks the mission completed (`src-tauri/src/commands/mission.rs:358-365`). `kill` only drops the PTY master (`src-tauri/src/session/manager.rs:266-277`); it has no child handle, no SIGKILL escalation, and no wait/poll for the session DB row to leave `running`. This can produce a completed mission while child processes are still alive or session rows still say `running`, which violates the C6 clean-shutdown contract.

3. Medium: short-lived commands can race the live-session map insert. The reader thread starts before the session is inserted into the in-memory map (`src-tauri/src/session/manager.rs:207-229` vs `manager.rs:233-242`). If the child exits quickly, `forget()` runs before the map entry exists, then `spawn` inserts a stale live handle that no thread will remove. `session_inject_stdin`/`session_kill` can then target a session that already exited.

4. Medium: C6 PATH wiring is missing. The implementation docs require spawn to prepend `$APPDATA/runners/bin` so the bundled `runners` CLI wins, but the spawn env setup only applies runner env plus `RUNNERS_*` vars (`src-tauri/src/session/manager.rs:149-164`). When C9 lands, agents may not find the intended CLI, or may call an unrelated system `runners`.

Validation: `cargo test --workspace` passed 57 tests, `pnpm exec tsc --noEmit` passed, `pnpm lint` passed, and `git diff --check origin/feature/v0-mvp...HEAD` is clean.

yicheng47:
Review findings:

1. High: `mission_start` can return an error after creating a running mission and spawning earlier sessions. It commits the mission first, then loops `spawn(...)?` for each runner (`src-tauri/src/commands/mission.rs:329-358`). If a later runner fails to spawn, earlier sessions keep running and the mission stays `running`, but the caller receives only an error. The command needs all-or-nothing cleanup/abort semantics, or an explicit successful partial-start recovery state.

2. High: `mission_stop` does not wait for sessions to terminate before marking the mission completed. It calls `kill_all_for_mission`, ignores the result, then immediately completes the mission (`src-tauri/src/commands/mission.rs:363-370`). `kill` just drops the PTY handle (`src-tauri/src/session/manager.rs:266-278`), with no child wait, DB-status wait, or escalation path. This can leave child processes or `sessions.status = running` after the mission row is already completed.

3. Medium: short-lived commands can race the live-session map insert. The reader thread starts at `src-tauri/src/session/manager.rs:207-229` and can call `forget` before the map insert at `manager.rs:232-242`, leaving a stale live handle for a session that has already exited. `session_inject_stdin` / `session_kill` can then target a dead session.

4. Medium: C6 does not prepend `$APPDATA/runners/bin` to `PATH`. Spawn sets runner env plus `RUNNERS_*` vars (`src-tauri/src/session/manager.rs:149-164`), but `docs/impls/v0-mvp.md:160` and `docs/arch/v0-arch.md:303` require the bundled `runners` CLI to be first on PATH. When C9 lands, agents may not find the intended CLI, or may invoke an unrelated system `runners`.

Validation: `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`, `pnpm exec tsc --noEmit`, `pnpm run lint`, and `git diff --check origin/feature/v0-mvp...HEAD` all passed.

yicheng47:
One remaining finding after the follow-up patch:

1. Medium: `SessionManager::spawn` can still leak a child on post-spawn setup failures. The process is launched at `src-tauri/src/session/manager.rs:188-191`, but the reader/writer setup and `sessions` insert happen afterward (`manager.rs:197-218`). If any of those fail, `spawn` returns `Err` before the session is in the live map, so the rollback in `mission_start` (`src-tauri/src/commands/mission.rs:359-372`) cannot kill it. Cleanup should kill/wait the child on every post-`spawn_command` error path, or move the fallible setup before launching where possible.

The four previously posted issues are fixed in the normal path. Validation passed: `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`, `pnpm exec tsc --noEmit`, `pnpm run lint`, and `git diff --check origin/feature/v0-mvp...HEAD`.

PR #13, opened 2026-04-23
feat(db,commands): shared runners + direct-session schema (C5.5a)

yicheng47:
## Summary

C5.5a of the v0 MVP umbrella — schema + backend for the two product
changes requested after C3 shipped:

1. **Runners are top-level and shared.** A runner is its own entity; a
   crew is a composition of existing runners via a new \`crew_runners\`
   join table. One runner can belong to many crews, and the handle is
   globally unique so \`@impl\` means the same runner everywhere it
   appears in the event log.
2. **Sessions without missions.** \`sessions.mission_id\` is now nullable
   so C6 can spawn a direct-chat PTY without firing a mission. Sessions
   carry their own \`cwd\` for that case.

Plan doc: [\`docs/impls/v0-mvp-c5-5-shared-runners.md\`](./docs/impls/v0-mvp-c5-5-shared-runners.md). Positioned between C5 (shipped) and C6 (PTY runtime — which needs the new sessions shape).

## What changed

**Schema rewrite — \`src-tauri/migrations/0001_init.sql\`** (MVP has no prod data, so we rewrite DDL in place rather than layering a 0002 migration):
- \`runners\` drops \`crew_id\`, \`position\`, \`lead\`. Adds \`UNIQUE(handle)\`.
- New \`crew_runners(crew_id, runner_id, position, lead, added_at)\` with the \`one_lead_per_crew\` partial unique index moved here.
- \`sessions.mission_id\` nullable (\`ON DELETE SET NULL\`); added \`sessions.cwd\`.

**Rust backend:**
- \`model.rs\`: new \`CrewSlot\` and \`CrewRunner\`; \`Runner\` loses slot fields; \`Session.{mission_id,cwd}\` become \`Option\`.
- \`commands/runner.rs\`: global CRUD + \`runner_activity(id)\` (active sessions / active missions / crew count / last started). Global delete auto-promotes in every crew where the runner was lead.
- **\`commands/crew_runner.rs\` (new):** \`list\`, \`add_runner\`, \`remove_runner\`, \`set_lead\`, \`reorder\` — the old lead-invariant logic moved off of \`runners\` onto the join. \`set_lead\` clears the old lead before setting the new so the partial unique index never transiently violates; \`reorder\` parks rows at negative positions to dodge \`UNIQUE(crew_id, position)\` during swaps.
- \`commands/mission.rs\`: roster read via \`crew_runner::list\` instead of \`runner::list(crew_id)\`.

**Frontend (kept compiling, full Runners browse page deferred to C5.5b):**
- \`api.ts\`: new \`api.crew.{listRunners,addRunner,removeRunner,setLead,reorder}\` + \`api.runner.activity\`. Runner CRUD drops \`crew_id\`.
- \`types.ts\`: new \`CrewRunner\`, \`RunnerActivity\`.
- \`CrewEditor.tsx\`: uses \`CrewRunner[]\` for the slot list; slot-row delete becomes \"Remove from crew\" (global runner delete will live on the Runners page).
- \`AddSlotModal.tsx\`: two-step create (\`runner_create\` then \`crew_add_runner\`) with rollback if the second step fails — avoids leaving orphaned global runners.

## Dev upgrade note

Delete local \`\$APPDATA/runners/runners.db\` once to pick up the new schema.

## Test plan

- [x] \`cargo test --workspace\` — 61 tests pass (16 new in crew_runner, 8 new in runner, updated db/crew/mission)
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\` clean
- [x] \`cargo fmt --all --check\` clean
- [x] \`pnpm run lint\` clean
- [ ] \`pnpm tauri dev\` smoke: create two crews, add the same runner to both (verify sharing), remove the shared runner from one crew (verify auto-promote), delete the shared runner globally (verify both crews auto-promote their next member)

## Follow-up (C5.5b, separate PR)

- \`/runners\` browse page with activity badges + global edit/delete.
- \"Add slot\" modal expanded into a picker (\"pick existing runner\" vs. \"create new\").
- Sidebar: re-add \`Runner\` between \`Crew\` and \`Mission\`.
- \"Chat with runner\" entrypoint will land alongside C6's PTY plumbing.


yicheng47:
Review findings:

1. Medium: `crew_runners.position` is documented as dense, but removal paths leave holes. `src-tauri/src/commands/crew_runner.rs:10` declares `0, 1, 2, ...`; however `remove_runner` deletes/promotes without renumbering (`crew_runner.rs:160-188`), and global `runner_delete` has the same issue after the cascade removes memberships (`src-tauri/src/commands/runner.rs:262-286`). Since `crew_add_runner` appends at `MAX(position)+1` (`crew_runner.rs:118-130`), a crew with positions `0,1,2` becomes `0,2`, then `0,2,3`. That leaks through the `CrewRunner.position` API and breaks the stated roster invariant. Renumber survivors inside the same transaction after membership removal/global delete, and add regression tests for removing middle/lead members and deleting a shared runner.

Validation: `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`, `pnpm exec tsc --noEmit`, `pnpm run lint`, and `git diff --check origin/feature/v0-mvp...HEAD` all passed.

PR #17, opened 2026-04-26
feature(c8): direct-chat frontend with xterm + sidebar SESSION list

yicheng47:
## Summary
- xterm.js direct-chat pane (`/runners/:handle/chat`) replacing the v0 `<pre>` debug surface — real TUIs (claude-code, codex) render with ANSI / cursor / mouse correctly, full-width
- Persistent sidebar with WORKSPACE nav and a live SESSION list driven by `runner/activity` events; click an active session to re-attach to the live PTY across reloads
- Backend plumbing for size-aware sessions: `session_resize`, `session_start_direct(cols, rows)`, plus startup reconciliation that flips orphaned `running` rows to `stopped` so post-restart navigation never lands on a dead PTY id

## Key changes
- **Layout-route AppShell.** Hoisted out of every page so the sidebar (and its event listener) mounts exactly once. Without this, navigating to the chat page tore down the listener, and the activity event from the spawn that immediately followed was lost.
- **`direct_session_id` on `RunnerActivity` / `runner/activity`.** Sidebar seeds and refreshes the in-memory `handleToSessionId` map from this, so SESSION clicks survive a window reload.
- **PTY size handshake.** Frontend fits xterm before calling `startDirect` and forwards `cols`/`rows`; later window resizes refit and call `session_resize`. Claude-code's input frame paints at the correct width on first render instead of stuck at 80×24.
- **Carbon & Plasma theme.** Tailwind 4 `@theme` tokens, dark scrollbars, Apple-safe-area macOS app icon, custom `RuntimeSelect`, and `EmptyStateCard` shared by Runners/Crews.

## Test plan
- [ ] `cargo check --lib` / `pnpm tsc --noEmit` clean (verified locally)
- [ ] `pnpm tauri dev` — Runners → Chat now → claude-code TUI renders at full width, keystrokes round-trip, ANSI colors correct
- [ ] Background the window, foreground it — TUI repaints (canvas refresh)
- [ ] Sidebar SESSION list shows the running runner with a green dot; clicking it stays on the same chat page (no remount blank)
- [ ] End chat → status reads "stopped", not "crashed"
- [ ] Reload the window — sidebar still shows the live session, clicking re-attaches
- [ ] Quit the app → relaunch — startup reconciliation flips the orphaned `running` row to `stopped`; sidebar shows "No live sessions"


yicheng47:
Review findings:

1. Medium: reload reattach is broken on the chat route itself. `src/pages/RunnerChat.tsx:272` rejects `/runners/:handle/chat` when `location.state` is missing, which is exactly what happens after a window reload. The sidebar does rehydrate `direct_session_id`, but `src/components/Sidebar.tsx:73` no-ops when already on the same chat path, so clicking the SESSION item cannot pass `{ sessionId }` back in. This contradicts the PR test plan’s reload/reattach flow.

2. Medium: the output/exit listeners are not actually registered before `session_start_direct`. `src/pages/RunnerChat.tsx:243` starts `listen(...)` asynchronously, but `src/pages/RunnerChat.tsx:334` invokes the spawn immediately. A fast command can emit its first output or exit before the Promise resolves, leaving the terminal missing startup bytes or stuck as `running`. The spawn path should run only after both unlisten functions are acquired.

Verification:
- `pnpm run lint` passed.
- `cargo test --workspace` passed, 84 tests.

PR #18, opened 2026-04-26
feature(c8): signal router v0 + runtime adapter for system_prompt

yicheng47:
## Summary
- Reframe C8 in the v0-mvp plan from "orchestrator v0" to "signal router v0" — a flat parent-process dispatcher, not a rule engine. The lead runner owns coordination judgment; C8 owns the plumbing the lead can't do from inside a child PTY (bootstrap, cross-process stdin push, UI bridge).
- Explicitly descope dispatch ledger / replay idempotence, inbox-summary enrichment, and the rule/policy abstraction. Each comes with a paragraph explaining why the lead handles the responsibility instead.
- Fold the previously-separate "Cross-cutting launch/prompt adapter" block into C8 — the prompt composer feeds the `mission_goal` handler directly.
- Sync `docs/tests/v0-mvp-tests.md`: drop the dispatch-ledger replay test and the `inject_stdin_*` audit-event tests, add `pending_ask_map_reconstructs_from_log_on_reopen`, add the runtime-adapter tests, and rewrite the `mission_e2e` driver pseudocode so stdin pushes are silent (no synthetic events for handler invocations).

## Why this changed

Previous plan's framing made C8 sound like a framework with a rule engine and audit-event ledger. In practice the only mechanisms MVP needs are five hardcoded `match` arms:

| Signal | Mechanism |
|---|---|
| `mission_goal` | Inject composed launch prompt to lead's stdin |
| `human_said` | Inject `payload.text` to `payload.target` (or lead) |
| `ask_lead` | Inject worker's `{question, context}` to lead's stdin |
| `ask_human` | Append `human_question` event for the workspace UI |
| `human_response` | Look up `question_id`, inject answer to original asker |

There's no policy to evolve and no rule to override, so the framework framing was buying complexity we'd never use. v0.x can revisit if user-defined signal types ship.

## Code path

No code changes in this PR — plan/test docs only. The implementation PR will rename `src-tauri/src/orchestrator/` → `src-tauri/src/router/` and ship the five handlers + the prompt composer + the runtime adapter (claude-code `--append-system-prompt` on both `spawn` and `spawn_direct`).

## Test plan
- [x] Plan and test docs are internally consistent (no orchestrator-as-engine references left except: historical reframing context, the actual SQL column name `crews.orchestrator_policy`, and the existing module-stub path being renamed)
- [ ] Reviewer agrees with the descope rationale before the implementation PR builds against this plan

### DIFF
--- a/docs/arch/v0-arch.md
+++ b/docs/arch/v0-arch.md
@@ -532,11 +532,11 @@ Stdin pushes are deliberately silent: the router writes bytes into the target PT
 ```jsonc
 // When the card is shown:
 {
+  "id": "01HG...",                             // canonical question_id (use this in human_response)
   "kind": "signal",
   "type": "human_question",
   "from": "router",
   "payload": {
-    "question_id": "01HG...",                  // = this event's id; echoed here for convenience
     "triggered_by": <triggering-signal.id>,    // e.g. the changes_requested signal's id
     "prompt": "Reviewer requested changes. Accept or override?",
     "choices": ["accept", "override"],
@@ -550,13 +550,15 @@ Stdin pushes are deliberately silent: the router writes bytes into the target PT
   "type": "human_response",
   "from": "human",
   "payload": {
-    "question_id": <human_question.id>,       // lets rules target a specific card
+    "question_id": <human_question.id>,       // = the card event's `id` field
     "choice": "accept"                         // the clicked value (always one of choices[])
   }
 }
 ```
 
-Causality is carried in-payload rather than on the envelope: `human_question.payload.triggered_by` records which signal opened the card, and `human_response.payload.question_id` records which card the human clicked. The router needs no additional schema fields to match them.
+Causality is carried in-payload rather than on the envelope: `human_question.payload.triggered_by` records which signal opened the card, and `human_response.payload.question_id` records which card the human clicked (set to the `human_question` event's own `id`). The router needs no additional schema fields to match them.
+
+**Implementation note.** `human_question` does *not* echo `question_id` into its own payload. The canonical id is the event's own `id` field, and the parent process can't know that id until after the flock-guarded log append assigns it (pre-allocating would break the cross-process monotonic-ULID invariant in §5.1.1). UI consumers should read `human_question.id` and pass it through as `human_response.payload.question_id`.
 
 If two `ask_human` prompts are ever outstanding at once and we need richer discrimination, v0.x will add event-DAG fields. v0 ships the simple case; concurrent prompts are out of scope.
 
--- a/docs/arch/v0-arch.md
+++ b/docs/arch/v0-arch.md
@@ -21,12 +21,12 @@ Runners is a local desktop app. A user configures a **crew** of CLI coding agent
 │           │                          │                          │           │
 │           │                          │                          ▼           │
 │           │                          │               ┌──────────────────┐   │
-│           │                          │               │ Orchestrator     │   │
-│           │                          │               │  - policy rules  │   │
-│           │                          │               │  - action dispch │   │
+│           │                          │               │ Signal router    │   │
+│           │                          │               │  - fixed handlers│   │
+│           │                          │               │  - status map    │   │
 │           │                          │               └────────┬─────────┘   │
 │           │                          │                        │             │
-│           │      inject_stdin / ask_human / pause / ...       │             │
+│           │      inject_stdin / human_question / status        │             │
 │           └─────────────────────────►│◄───────────────────────┘             │
 │                                      ▼                                      │
 │  ┌──────────────────────────────────────────────────────────────────────┐   │
@@ -44,7 +44,7 @@ Runners is a local desktop app. A user configures a **crew** of CLI coding agent
 │                  ┌─────────────────────────────┐                          │
 │                  │  events.ndjson (per mission)│                          │
 │                  └──────────────┬──────────────┘                          │
-│                                 │ notify → EventBus → Orchestrator + UI   │
+│                                 │ notify → EventBus → Router + UI         │
 └─────────────────────────────────┼─────────────────────────────────────────┘
                                   ▼
                          ┌───────────────────────┐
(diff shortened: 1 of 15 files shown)


yicheng47:
**Code Review Findings**

- **P1:** `src-tauri/src/commands/mission.rs:419` mounts the bus before sessions are spawned/registered, but `src-tauri/src/event_bus/mod.rs:180` starts replay immediately on a consumer thread. If `mission_goal` is replayed before `router.register_sessions(...)` at `mission.rs:476`, `inject_to_handle` sees no session and only emits a warning. The lead may never receive the launch prompt. Mount after spawn + registration, or make router replay wait until sessions are registered.

- **P1:** `src-tauri/src/router/runtime.rs:31` adds `--instructions` for `codex`, but the installed Codex CLI rejects it with `error: unexpected argument --instructions found`. Any Codex runner with `system_prompt` will fail to start. Until the correct Codex prompt mechanism is verified, omit the flag or gate it behind a known-supported version.

- **P2:** `src-tauri/src/router/handlers.rs:123-124` uses the same live handler for replay reconstruction and unconditionally appends a new `human_question`. On router remount/replay, historical `ask_human` events will duplicate cards; historical `human_response` / `runner_status idle` can also re-inject old stdin. The router needs separate projection replay from live side effects.

- **P2:** `src-tauri/src/router/mod.rs:255` sets `human_question.payload.question_id` to the triggering `ask_human` id, while `docs/arch/v0-arch.md:539` says it equals the `human_question` event id. If C10 follows the architecture and sends the card id, `handlers.rs:136` will treat the response as unknown. Align the protocol now, either in code or docs/tests.

Checks run:
- `pnpm exec tsc --noEmit` passed under Node v22.22.2.
- `pnpm run lint` passed under Node v22.22.2.
- Worktree was clean after verification.

yicheng47:
Thanks for the review. Plan to address all four:

**P1 — bus-mount/registration race.** Move bus mount to *after* the spawn loop and `register_sessions`. The bus replays from offset 0 either way, so the opening `mission_start` / `mission_goal` events still surface; runners haven't written anything yet at that point because their PTYs only come up during the spawn loop. This closes the gap without buffering. (The original "mount before spawn" comment in `mission.rs` was guarding against runner startup writes — moving mount after spawn doesn't lose anything because the bus still does an initial replay from byte 0.)

**P1 — `codex --instructions` is wrong.** Drop the codex branch in `router/runtime.rs`; treat unknown runtimes as no-flag (with a code comment explaining the gap). Will revisit when there's a verified Codex prompt mechanism.

**P2 — replay vs live side effects.** Add `Router::reconstruct_from_log()` that folds `ask_human` / `human_question` / `human_response` / `runner_status` into the pending-ask + runner-status maps without injecting stdin or appending events. Capture the end-of-log ULID as a replay high-water mark; live `handle_event` short-circuits events whose id is ≤ that mark. Reconstruction runs before the bus mounts, so the bus's initial replay still hits the dispatcher but no-ops past the watermark and only acts on genuinely new events.

**P2 — `question_id` schema mismatch.** Align with arch §5.5.0: `question_id == human_question.id`, not `ask_human.id`. The pending-ask map keys on the appended `human_question` event id. `ask_human` handler becomes: append `human_question` → capture the assigned id → record `human_question.id → asker`. Replay walks the log pairing each `ask_human` with its following `human_question` via `human_question.payload.triggered_by`, so the map reconstructs without needing extra schema fields.

Will push the fixes as a follow-up commit on the same branch.

yicheng47:
Plan looks right overall. One caveat on the replay high-water approach:

For a **fresh `mission_start`**, do not run reconstruction / set the replay high-water over the just-written opening events before the bus mounts. If `mission_goal` is `<= high_water`, the bus initial replay will no-op it and the lead still will not receive the launch prompt.

I would split the modes explicitly:

- Fresh start: spawn sessions → `register_sessions` → mount bus with no historical suppression, so the initial replay from offset 0 delivers `mission_start` / `mission_goal` and bootstraps the lead.
- Reopen/remount: reconstruct projection state from historical `ask_human` / `human_question` / `human_response` / `runner_status`, record a high-water mark, then mount bus and suppress side effects for events at or below that mark.

The other three parts look aligned: dropping the unverified Codex flag, separating projection replay from live side effects, and keying pending asks by `human_question.id` while using `triggered_by` to recover the original `ask_human`/asker relationship.

yicheng47:
Pushed d9968b5 with all four fixes:

**P1 — bus mount race (`commands/mission.rs`).** Reordered: open log → build router → spawn sessions → `register_sessions` → mount bus. The bus's initial replay now lands on a fully-wired handle map, so `mission_goal` reliably reaches the lead. Spawn-before-mount is safe because runners can't write to the log until they get their first stdin (which only arrives after `mission_goal` is delivered post-mount).

**P1 — codex flag (`router/runtime.rs`).** Dropped the `--instructions` branch; Codex falls through to no-flag with a comment explaining the gap. Test renamed to `codex_runtime_omits_flag_until_verified_mechanism`.

**P2 — replay vs live (`router/mod.rs`).** Added `Router::reconstruct_from_log()` (reopen-only) that folds `pending_asks` + `runner_status` without firing handlers, then sets a replay high-water ULID. `handle_event` short-circuits events with `id ≤ watermark`, so the bus's initial replay no longer re-emits cards or re-injects historical stdin. **`mission_start` does NOT call `reconstruct`** — setting the watermark over the just-written `mission_goal` would suppress the bootstrap (your caveat). Two new tests cover this: `reconstruct_recovers_latest_runner_status_only` and `fresh_mission_start_does_not_call_reconstruct_so_mission_goal_fires`.

**P2 — `question_id` schema (`router/handlers.rs` + `router/mod.rs`).** `append_human_question` now returns the appended event's id; `ask_human` handler keys `pending_asks` on the card id (the canonical `question_id` per arch §5.5.0). Reconstruct pairs each `ask_human` with its following `human_question` via `payload.triggered_by`. The convenience-echo `payload.question_id` is omitted because the id isn't known until after append — consumers should read `event.id`. Existing tests updated to reference the card id.

Checks:
- `cargo test --workspace` — 85 passed (up from 83 with two new replay tests).
- `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `pnpm exec tsc --noEmit` + `pnpm run lint` clean.

yicheng47:
**Follow-up Review Findings**

- **P1:** Backend CI is still red because `cargo fmt --all --check` fails. Example at `src-tauri/src/commands/mission.rs:462`, plus router/test/session formatting diffs. Run `cargo fmt --all` and push.

- **P2:** `human_question.payload.question_id` is now omitted in `src-tauri/src/router/mod.rs:357`, but the architecture still documents it as present at `docs/arch/v0-arch.md:539`. Either update the docs/tests contract or include the field, otherwise C10 may implement against stale protocol.

- **P3:** `reconstruct_from_log()` uses strict `read_from(0)` at `src-tauri/src/router/mod.rs:195`, while the bus uses lossy reads to skip malformed lines. That can make reopen reconstruction fail on a log the bus can otherwise tolerate.

Checks run locally:
- `cargo clippy --workspace --all-targets -- -D warnings` passed.
- `pnpm exec tsc --noEmit` passed under Node v22.22.2.
- `pnpm run lint` passed under Node v22.22.2.
- `cargo fmt --all --check` failed.
- Worktree was clean after verification.

yicheng47:
Pushed 17c164f addressing all three follow-ups:

**P1 — `cargo fmt --all --check`.** Ran the formatter across the C8 changes; CI should now go green. No behavior delta.

**P2 — `human_question.payload.question_id` doc/code drift.** Updated `docs/arch/v0-arch.md` §5.5.0 to drop the echo and added an Implementation note: the canonical `question_id` is the event's own `id` field (the parent process can't pre-allocate the id without breaking the flock-guarded monotonic-ULID invariant in §5.1.1). Only `human_response.payload.question_id` carries it. UI consumers read `human_question.id` and pass it through.

**P3 — `reconstruct_from_log` strict read.** Switched to `read_from_lossy(0)`; skipped offsets are logged via the same eprintln pattern as `event_bus`. Added `reconstruct_tolerates_malformed_lines_like_the_bus`: pre-seeds `ask_human` → bad NDJSON line → `human_question` (via live router), then on reopen reconstruct must still recover the pending-ask mapping. Also switched the test-only `read_signals` helper to lossy so the corrupted-log assertion can still inspect signals.

Checks:
- `cargo fmt --all --check` clean.
- `cargo test --workspace` — 86 passed (up from 85).
- `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `pnpm exec tsc --noEmit` + `pnpm run lint` clean.

PR #19, opened 2026-04-27
feature(c9): runner CLI binary

yicheng47:
Implements C9 from \`docs/impls/v0-mvp.md\`: the standalone \`runner\` binary that spawned agents use to append to the mission's NDJSON event log. Fresh chunk; the implementation plan lives in the first PR comment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,6 +54,56 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
+version = "1.0.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
+
+[[package]]
+name = "anstyle-parse"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "52ce7f38b242319f7cabaa6813055467063ecdc9d355bbb4ce0c68908cd8130e"
+dependencies = [
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle-query"
+version = "1.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
+dependencies = [
+ "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "anstyle-wincon"
+version = "3.0.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
+dependencies = [
+ "anstyle",
+ "once_cell_polyfill",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -480,6 +530,52 @@ dependencies = [
  "wasm-bindgen",
  "windows-link 0.2.1",
 ]
+
+[[package]]
+name = "clap"
+version = "4.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1ddb117e43bbf7dacf0a4190fef4d345b9bad68dfc649cb349e7d17d28428e51"
+dependencies = [
+ "clap_builder",
+ "clap_derive",
+]
+
+[[package]]
+name = "clap_builder"
+version = "4.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "714a53001bf66416adb0e2ef5ac857140e7dc3a0c48fb28b2f10762fc4b5069f"
+dependencies = [
+ "anstream",
+ "anstyle",
+ "clap_lex",
+ "strsim",
+]
+
+[[package]]
+name = "clap_derive"
+version = "4.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f2ce8604710f6733aa641a2b3731eaa1e8b3d9973d5e3565da11800813f997a9"
+dependencies = [
+ "heck 0.5.0",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
+name = "clap_lex"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c8d4a3bb8b1e0c1050499d1815f5ab16d04f0959b233085fb31653fbfc9d98f9"
+
+[[package]]
+name = "colorchoice"
+version = "1.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1d07550c9036bf2ae0c684c4297d503f838287c83c53686d05370d0e139ae570"
 
 [[package]]
 name = "combine"
@@ -1943,6 +2039,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "is_terminal_polyfill"
+version = "1.70.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a6cb138bb79a146c1bd460005623e142ef0181e3d0219cb493e02f7d08a35695"
+
+[[package]]
 name = "itoa"
 version = "1.0.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2567,6 +2669,12 @@ name = "once_cell"
 version = "1.21.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f7c3e4beb33f85d45ae3e3a1792185706c8e16d043238c593331cc7cd313b50"
+
+[[package]]
+name = "once_cell_polyfill"
+version = "1.70.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "384b8ab6d37215f3c5301a95a4accb5d64aa607f1fcb26a11b5303878451b4fe"
 
 [[package]]
 name = "open"
@@ -3408,6 +3516,19 @@ dependencies = [
  "tauri-plugin-shell",
  "tempfile",
  "thiserror 1.0.69",
+ "ulid",
+]
+
+[[package]]
+name = "runner-cli"
+version = "0.1.0"
+dependencies = [
+ "chrono",
+ "clap",
+ "runner-core",
+ "serde",
+ "serde_json",
+ "tempfile",
  "ulid",
 ]
 
@@ -4979,6 +5100,12 @@ name = "utf8_iter"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6c140620e7ffbb22c2dee59cafe6084a59b5ffc27a8859a5f0d494b5d52b6be"
+
+[[package]]
+name = "utf8parse"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "06abde3611657adf66d383f00b093d7faecc7fa57071cce2578660c9f1010821"
 
 [[package]]
 name = "uuid"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ resolver = "2"
 members = [
     "src-tauri",
     "crates/runner-core",
+    "cli",
 ]
 
 [workspace.package]
@@ -11,6 +12,7 @@ authors = ["Jason Wang <yuesihan@gmail.com>"]
 
 [workspace.dependencies]
 chrono = { version = "0.4", features = ["serde"] }
+clap = { version = "4", features = ["derive"] }
 fs2 = "0.4"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "runner-cli"
+version = "0.1.0"
+edition.workspace = true
+authors.workspace = true
+description = "Bundled `runner` CLI agents use to append signals/messages to a mission's NDJSON event log."
+
+# Bin name = "runner" so the on-disk file matches the PATH expectation in
+# `SessionManager::spawn` / `spawn_direct` ($APPDATA/runner/bin/runner).
+[[bin]]
+name = "runner"
+path = "src/main.rs"
+
+[dependencies]
+runner-core = { path = "../crates/runner-core" }
+clap = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+chrono = { workspace = true }
+ulid = { workspace = true }
+
+[dev-dependencies]
+tempfile = "3"
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -5,10 +5,13 @@ edition.workspace = true
 authors.workspace = true
 description = "Bundled `runner` CLI agents use to append signals/messages to a mission's NDJSON event log."
 
-# Bin name = "runner" so the on-disk file matches the PATH expectation in
-# `SessionManager::spawn` / `spawn_direct` ($APPDATA/runner/bin/runner).
+# Source-side bin name is `runner-cli` so it doesn't collide with the
+# Tauri app crate, which also produces a `runner` binary in the same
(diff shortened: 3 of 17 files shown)


yicheng47:
## C9 — `runner` CLI binary: implementation plan

### Goal

A standalone \`runner\` binary that agents inside spawned PTYs invoke to:
- emit signals (\`runner signal <type> [--payload <json>]\`)
- post messages (\`runner msg post <text> [--to <handle>]\`)
- read their inbox (\`runner msg read [--since <ulid>] [--from <handle>]\`)
- report availability (\`runner status busy|idle [--note <text>]\`, sugar around \`signal runner_status\`)
- print help (\`runner help\`)

The binary is installed at \`\$APPDATA/runner/bin/runner\` so the existing PATH prepend in \`SessionManager::spawn\` wins (arch §5.3 Layer 2). Tracked in \`docs/impls/v0-mvp.md#C9\` and tested per \`docs/tests/v0-mvp-tests.md#I2\`.

### Scope of this PR

- Add a new \`cli/\` workspace member producing a binary called \`runner\`.
- Implement the five verbs above plus help.
- Wire the existing \`runner_core::event_log\` for append/read — no log code duplicated.
- Validate envelope inputs: signal type must be in the per-crew allowlist (\`signal_types.json\` sidecar), \`--to\` handle must exist in the same crew (covered in I2.4 — needs a roster sidecar; see "Open question" below).
- Install the built binary into \`\$APPDATA/runner/bin\` at app launch (Tauri setup hook), and wire a dev-mode fallback for \`cargo test\` driving the CLI directly.
- Direct-chat sessions (no \`RUNNER_*\` env vars) — every verb either no-ops cleanly with a clear stderr message or prints help, never crashes the agent process.

Out of scope: the actual app-level "install at first launch" plumbing if it touches release packaging beyond a copy-on-startup. Will land minimal copy-on-startup logic and flag follow-ups if packaging is more involved.

### Crate layout

\`\`\`
cli/
├─ Cargo.toml              # binary "runner", deps: runner-core, clap, serde, serde_json, anyhow
├─ src/
│  ├─ main.rs              # arg parsing → dispatch
│  ├─ env.rs               # resolve RUNNER_CREW_ID / MISSION_ID / HANDLE / EVENT_LOG; clear no-bus errors
│  ├─ allowlist.rs         # read \`\$APPDATA/runner/crews/{id}/signal_types.json\`
│  ├─ signal.rs            # \`runner signal\` (+ \`runner status\` wrapper)
│  ├─ msg.rs               # \`runner msg post\` and \`runner msg read\` (+ inbox_read emission)
│  └─ help.rs              # \`runner help\` text
└─ tests/
   └─ roundtrip.rs         # I2 integration tests: spawn the built binary, assert log contents
\`\`\`

Workspace \`Cargo.toml\` gains \`cli\` as a member. The binary name is \`runner\` (set via \`[[bin]] name = "runner"\` so the produced file matches the PATH expectation regardless of the crate folder name).

### Verb behavior

**\`runner signal <type> [--payload <json>]\`** — appends a single \`EventDraft::signal\` to the log. Validates \`<type>\` against the sidecar allowlist; rejects unknown types with a stderr pointer at the allowlist (I2.2). \`--payload\` is parsed as JSON; default is \`{}\`. The CLI does not enrich payloads — it's a thin shell over \`EventLog::append\`.

**\`runner msg post <text> [--to <handle>]\`** — appends an \`EventDraft::message\`. \`--to\` is optional; absent means broadcast. The roster check (I2.4) needs the recipient list — either we read \`crew_runners\` somehow (CLI doesn't have DB access; cross-cutting concern), OR we ship a parallel \`\$APPDATA/runner/crews/{id}/roster.json\` sidecar that \`mission_start\` writes alongside \`signal_types.json\`. **Recommendation:** add a roster sidecar in this PR — same pattern as the signal-types one, atomic write via \`tempfile::persist\`. Cheap, keeps CLI DB-free.

**\`runner msg read [--since <ulid>] [--from <handle>]\`** — projects the caller's inbox: \`kind = "message" AND (to == null OR to == \$RUNNER_HANDLE)\`. Filters: \`--since\` lower-bounds by ULID, \`--from\` filters by sender. Prints in append order (one block per message, same shape as the workspace UI feed). Then appends one \`signal inbox_read\` with \`payload.up_to = <max ULID printed>\`, satisfying the per-runner watermark contract from C7.

**\`runner status busy|idle [--note <text>]\`** — a sugar wrapper that emits \`signal runner_status\` with \`payload = { state, note? }\`. Arch §5.5.1 + the C8 router handler already consume these.

**\`runner help\`** — prints a short summary with examples drawn straight from \`docs/arch/v0-arch.md\` §6.3. \`--help\` on each subcommand prints the same shape; clap gives this for free.

### Env contract

Required for every verb except \`help\`:
- \`RUNNER_CREW_ID\` — crew ULID, used to resolve the signal-type and roster sidecars.
- \`RUNNER_MISSION_ID\` — mission ULID, written into every appended event's envelope.
- \`RUNNER_HANDLE\` — the caller's handle, written into envelope \`from\`.
- \`RUNNER_EVENT_LOG\` — absolute path to the mission's \`events.ndjson\` (already exported by C6's \`SessionManager::spawn\`).

Missing env handling:
- **Mission session** (all four set) → normal behavior.
- **Direct-chat session** (none set) → exit 0 with a one-liner on stderr explaining the verb is a no-op outside a mission. This stops a curious agent from crashing on \`runner status idle\` in a direct chat (the C8.5 risk note explicitly calls this out).
- **Partially set** (e.g. \`RUNNER_HANDLE\` missing but \`RUNNER_EVENT_LOG\` present) → exit non-zero with a precise pointer at which var is missing. I2.8.

### Concurrency & atomicity

\`EventLog::append\` already takes the flock and assigns monotonic ULIDs cross-process (C4 contract). The CLI calls it once per verb invocation; no extra coordination needed. I2.7 (10 shells × 100 invocations each → exactly 1000 lines) is a straight test of that contract surfacing through the CLI.

### Distribution

At app startup, copy the built \`runner\` binary into \`\$APPDATA/runner/bin/\` (chmod +x on Unix). Source path: \`std::env::current_exe()\` → resolve sibling \`runner\` next to the Tauri binary in release; in dev (\`cargo run\`), the \`target/debug/runner\` path. Skip the copy if the destination already matches the source bytes (cheap mtime+size check) so launches stay fast.

For the I2 integration tests, the test fixture builds the CLI via \`cargo build -p cli\` and adds \`target/debug\` to PATH for the spawned shell — no installer dance needed in tests.

### Tests

- Unit (in \`cli/src/\`): allowlist parse, env resolution, payload JSON parse, inbox projection logic.
- Integration (\`cli/tests/roundtrip.rs\`) — covers I2.1 through I2.8 from the test plan: signal roundtrip, unknown signal rejection, msg routing, unknown handle rejection, msg read + inbox_read emission, runner_status emission, concurrent-writer atomicity, missing-env failure modes.

### Risks

- **Roster sidecar adds a write to \`mission_start\`.** Mirrors the signal-types sidecar. Atomic write is the same dance as \`write_signal_types_sidecar\`; trivial.
- **Packaging.** App-startup copy assumes the binary is built and present alongside the Tauri binary in production. If Tauri's bundler doesn't carry it into the app, follow-up needed for prod packaging. This PR ships the dev path; prod packaging stays a tracked open item.
- **clap vs hand-rolled parser.** clap is in the dep tree elsewhere (Tauri's transitive deps). Adding it here is fine; alternative would be a small hand-rolled match. Going with clap for ergonomic --help output.

### Order of work

1. Workspace \`cli/\` crate scaffold + bin name + dep wiring.
2. \`env.rs\` + \`allowlist.rs\` with unit tests.
3. \`signal\` verb (+ \`status\` wrapper) — covers the simplest end-to-end path.
4. \`msg post\` (with roster sidecar added to \`mission_start\`).
5. \`msg read\` + \`inbox_read\` emission.
6. \`help\`.
7. App-startup copy into \`\$APPDATA/runner/bin\` (Tauri setup hook).
8. I2 integration tests.
9. \`cargo fmt\` + clippy + tsc + lint.

### Open question

Roster sidecar — confirm the approach: write \`\$APPDATA/runner/crews/{crew_id}/roster.json\` alongside \`signal_types.json\` in \`mission_start\`. Contains \`[{ "handle": "...", "lead": true|false }, ...]\`. Used by \`runner msg post --to\` validation. Alternative is a per-mission sidecar (\`missions/{id}/roster.json\`) — probably cleaner since rosters are mission-scoped, not crew-scoped, even though crew membership is the source of truth. Will go with the per-mission location unless you disagree.

yicheng47:
No code to review yet: this draft PR currently has one empty placeholder commit and 0 changed files.

For the open question in the implementation plan: use the **per-mission roster sidecar**:

`$APPDATA/runner/crews/{crew_id}/missions/{mission_id}/roster.json`

Reason: roster validity is mission-scoped once the mission starts. If crew membership changes later, a running or historical mission should still validate `runner msg post --to <handle>` against the frozen roster from that mission, not the crew’s current membership. The CLI can locate the sidecar from `RUNNER_EVENT_LOG`’s parent directory, so it does not need DB access or an extra app-data env var.

yicheng47:
Pushed 56461fa with the C9 implementation. Marked PR ready for review.

**What landed**

- New \`cli/\` workspace member; \`runner\` binary via clap.
- \`runner signal\` / \`status\` / \`msg post\` / \`msg read\` / \`help\` per arch §6.3.
- Per-mission roster sidecar (\`roster.json\` next to \`events.ndjson\`), written by \`mission_start\` via \`write_roster_sidecar\`. Uses the per-mission location per your guidance — frozen at mission start so historical \`--to\` lookups stay valid even if crew membership changes.
- App-startup CLI install (\`src-tauri/src/cli_install.rs\`) copies the built \`runner\` next to the Tauri exe into \`\$APPDATA/runner/bin/\`. Skip-if-current (size + mtime); chmod 755 on Unix; atomic via tempfile persist.
- Off-bus / partial-env handling per the plan: none-set → exit 0 + stderr notice (so direct-chat agents don't crash); partial → exit 2 with named missing var(s).
- \`msg read\` empty inbox does NOT emit \`inbox_read\` (avoids redundant log noise; the bus would de-dupe but skipping keeps the log readable).

**Tests**

- 11 CLI unit tests + 12 integration tests in \`cli/tests/roundtrip.rs\` cover I2.1–I2.8, plus extras: \`status\` state rejection, empty-inbox no-op, off-bus notice, help.
- Concurrent-writer test scaled down from spec's 10×100 to 5×20 (= 100 lines) — process-spawn overhead made the original prohibitive in CI without changing what's being verified (the flock contract). All ULIDs strictly monotonic; every line parses.
- 86 backend tests still passing.
- \`cargo fmt --all --check\`, \`cargo clippy --workspace --all-targets -- -D warnings\`, \`pnpm exec tsc --noEmit\`, \`pnpm run lint\` all clean.

**Notes / open items**

- The clap-derive \`Help\` enum variant collides with clap's auto-generated \`help\` subcommand; resolved by setting \`disable_help_subcommand = true\` on the top-level command. Documented in code.
- Production packaging (Tauri bundling \`runner\` next to the main exe in release builds) is not yet wired — the current install path assumes the binary is built and present alongside the Tauri exe. Dev path works end-to-end; prod packaging is a follow-up tracked here.
- C9's "msg read with --since" semantics: I implemented \`since\` as a strict lower bound (event id strictly greater than \`--since\`). This matches typical "fetch since last seen ULID" usage and lets the next read pick up exactly where the previous left off without overlap. Flag if you'd prefer inclusive.

yicheng47:
Reviewing the implementation, I found two blocking issues:

1. `src-tauri/src/cli_install.rs:59` - the bundled CLI likely never installs. The installer looks for a sibling `runner`, but the Tauri app binary is also named `runner`, so in dev the candidate is the current exe and gets skipped. `src-tauri/tauri.conf.json` also does not declare any external binary, so packaged builds do not appear to ship a separate CLI sidecar either. That leaves `$APPDATA/runner/bin` empty, and spawned agents will fail when they try to call `runner ...`.

2. `cli/src/msg.rs:118` - `runner msg read --from <handle>` filters which messages are printed, but still emits `inbox_read` with `up_to` set to the last printed message id. The event bus treats `inbox_read` as a global per-runner watermark and marks every inbox message `<= up_to` as read, regardless of sender, so a filtered read can silently clear unread messages from other senders that were never shown. Either suppress watermark emission for filtered reads or only advance to a watermark that is safe for the full inbox projection.

Local verification passed on PR head `56461faa`: `cargo fmt --all --check`, `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `pnpm exec tsc --noEmit`, and `pnpm run lint` under Node `22.22.2`.

yicheng47:
Pushed 581c267 with both blocker fixes.

**B1 — CLI binary name collision.** Confirmed root cause: the Tauri app crate's binary is also \`runner\`, and Cargo writes both to the same \`target/<profile>/\` dir, so they overwrite each other and \`cli_install\`'s sibling search finds only the Tauri binary (filtered out by the equality guard).

Fix:
- \`cli/Cargo.toml\`: \`[[bin]] name = "runner-cli"\`. Both binaries now coexist (\`target/debug/runner\`, \`target/debug/runner-cli\`).
- \`cli_install.rs\` looks for \`runner-cli\` next to \`current_exe()\` and copies it to \`\$APPDATA/runner/bin/runner\` (renaming the file on copy, so the on-disk PATH name stays \`runner\`).
- Updated integration tests to use \`CARGO_BIN_EXE_runner-cli\`.
- Production-bundle wiring (\`tauri.conf.json\` \`bundle.externalBin\` + a beforeBuildCommand step staging the binary at the target-triple-suffixed path the bundler expects) stays a tracked follow-up. The dev path now works end-to-end; v0 demos run in dev.

**B2 — \`--from\` watermark corruption.** Emitting \`inbox_read\` after a sender-filtered read advanced the bus's global per-runner watermark, silently clearing unread messages from other senders below \`up_to\`.

Fix: skip the \`inbox_read\` emission when \`--from\` is in effect. \`--since\` stays safe — it's a forward filter on an already-watermark-ordered projection, so the printed last id remains a sound "everything ≤ here has been seen" upper bound. Regression test \`i2_5c_msg_read_with_from_filter_does_not_emit_inbox_read\` seeds a mixed-sender inbox, runs the filtered read, and asserts no \`inbox_read\` was appended.

Checks:
- \`cargo test --workspace\` — 13 CLI integration tests + 11 CLI unit + 87 backend + 20 runner-core, all passing.
- \`cargo fmt --all --check\` clean.
- \`cargo clippy --workspace --all-targets -- -D warnings\` clean.
- \`pnpm exec tsc --noEmit\` + \`pnpm run lint\` clean.

Open follow-ups (tracked separately):
- Production sidecar packaging via \`bundle.externalBin\`. v0 ships in dev; this is a release-engineering chore that doesn't affect the demo path.

yicheng47:
Re-reviewed `581c267`. The `msg read --from` watermark issue is fixed, but the CLI install path still has a blocking gap.

`src-tauri/src/cli_install.rs:49` now looks for a sibling `runner-cli`, which avoids the binary-name collision, but a fresh Tauri app build does not produce that sibling. The documented dev path is `npm run tauri dev`, and `src-tauri/tauri.conf.json` only starts Vite before Tauri builds the app; it does not build the separate `runner-cli` workspace binary. Since `runner-cli` is not a dependency/target of the Tauri app package, `locate_source()` skips install unless someone manually ran `cargo build -p runner-cli` first. That still leaves `$APPDATA/runner/bin/runner` absent for a clean dev/demo setup, so spawned agents fail when they call `runner ...`.

Repro:

```sh
cargo build -p runner --bin runner --target-dir /tmp/runner-pr19-target
ls -l /tmp/runner-pr19-target/debug/runner /tmp/runner-pr19-target/debug/runner-cli
```

Result: `/tmp/runner-pr19-target/debug/runner` exists, `/tmp/runner-pr19-target/debug/runner-cli` does not.

Fix direction: wire the CLI build into the normal app dev/build path, for example by making the Tauri `beforeDevCommand` / `beforeBuildCommand` build or stage `runner-cli`, or by adding explicit packaging/staging support that guarantees the sibling exists before `install_runner_cli()` runs.

Local verification on `581c267` otherwise passed: `cargo fmt --all --check`, `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `pnpm exec tsc --noEmit`, and `pnpm run lint` under Node `22.22.2`.

yicheng47:
Pushed 8c3dc6d. The dev install path is now coupled to \`tauri dev\` itself:

\`src-tauri/tauri.conf.json\` now points \`beforeDevCommand\` / \`beforeBuildCommand\` at two new npm wrappers in \`package.json\`:

- \`tauri:before:dev\`: \`cargo build -p runner-cli && npm run dev\`
- \`tauri:before:build\`: \`cargo build -p runner-cli --release && npm run build\`

Order on \`tauri dev\`:
1. Tauri runs \`tauri:before:dev\` → \`cargo build -p runner-cli\` produces \`target/debug/runner-cli\`; then Vite starts (long-running).
2. Tauri builds the app: \`cargo build -p runner\` produces \`target/debug/runner\` next to it.
3. Tauri launches the app; \`cli_install\` finds the sibling \`runner-cli\` and copies it to \`\$APPDATA/runner/bin/runner\`.

Plain \`npm run dev\` is unchanged — the CLI build only fires when launching via Tauri, since that's the only path that consumes the sibling artifact.

Production-sidecar packaging (Tauri \`bundle.externalBin\` declaration so installer builds carry the CLI) remains a tracked follow-up; the dev demo path now works end-to-end with no manual \`cargo build -p runner-cli\`.

Verified locally: cargo build -p runner-cli produces \`target/debug/runner-cli\`; \`cargo test --workspace\`, \`cargo fmt --all --check\`, \`cargo clippy --workspace --all-targets -- -D warnings\`, \`pnpm exec tsc --noEmit\`, \`pnpm run lint\` all clean. (Did not run \`tauri dev\` itself because launching a dev server in this session needs explicit user permission.)

PR #20, opened 2026-04-27
feature(c10): mission workspace UI

yicheng47:
Implements C10 from \`docs/impls/v0-mvp.md\`: the \`/missions/:id\` workspace pane that surfaces the live mission — event feed, ask-human cards, runner rail, and embedded xterms — and reattaches cleanly on reopen. See the first PR comment for the implementation plan.

### DIFF
--- a/design/runners-design.pen
+++ b/design/runners-design.pen
@@ -3617,28 +3617,22 @@
                             },
                             {
                               "type": "frame",
-                              "id": "KzWib",
-                              "name": "slot1_r",
-                              "gap": 12,
+                              "id": "CEGkx",
+                              "name": "slot1_kebab",
+                              "width": 28,
+                              "height": 28,
+                              "cornerRadius": 6,
+                              "justifyContent": "center",
                               "alignItems": "center",
                               "children": [
                                 {
-                                  "type": "text",
-                                  "id": "3OtAa",
-                                  "fill": "#00FF9C",
-                                  "content": "Prompt",
-                                  "fontFamily": "Inter",
-                                  "fontSize": 12,
-                                  "fontWeight": "normal"
-                                },
-                                {
-                                  "type": "text",
-                                  "id": "QjvN7",
-                                  "fill": "#FF4D6D",
-                                  "content": "Remove",
-                                  "fontFamily": "Inter",
-                                  "fontSize": 12,
-                                  "fontWeight": "normal"
+                                  "type": "icon_font",
+                                  "id": "y1dheV",
+                                  "width": 16,
+                                  "height": 16,
+                                  "iconFontName": "ellipsis",
+                                  "iconFontFamily": "lucide",
+                                  "fill": "#9A9BA5"
                                 }
                               ]
                             }
@@ -3786,38 +3780,23 @@
                             },
                             {
                               "type": "frame",
-                              "id": "kc00i",
-                              "name": "slot2_r",
-                              "gap": 12,
+                              "id": "r7CLLj",
+                              "name": "slot2_kebab",
+                              "width": 28,
+                              "height": 28,
+                              "fill": "#1F2127",
+                              "cornerRadius": 6,
+                              "justifyContent": "center",
                               "alignItems": "center",
                               "children": [
                                 {
-                                  "type": "text",
-                                  "id": "P93jM",
-                                  "name": "set_lead_2",
-                                  "fill": "#9A9BA5",
-                                  "content": "Set as lead",
-                                  "fontFamily": "Inter",
-                                  "fontSize": 12,
-                                  "fontWeight": "normal"
-                                },
-                                {
-                                  "type": "text",
-                                  "id": "cKnvt",
-                                  "fill": "#00FF9C",
-                                  "content": "Prompt",
-                                  "fontFamily": "Inter",
-                                  "fontSize": 12,
-                                  "fontWeight": "normal"
-                                },
-                                {
-                                  "type": "text",
-                                  "id": "rEgQ0",
-                                  "fill": "#FF4D6D",
-                                  "content": "Remove",
-                                  "fontFamily": "Inter",
-                                  "fontSize": 12,
-                                  "fontWeight": "normal"
+                                  "type": "icon_font",
+                                  "id": "C4ExN",
+                                  "width": 16,
+                                  "height": 16,
+                                  "iconFontName": "ellipsis",
+                                  "iconFontFamily": "lucide",
+                                  "fill": "#EDEDF0"
                                 }
                               ]
                             }
@@ -3965,38 +3944,22 @@
                             },
                             {
                               "type": "frame",
-                              "id": "nFVw0",
-                              "name": "slot3_r",
-                              "gap": 12,
+                              "id": "C0lDY",
+                              "name": "slot3_kebab",
+                              "width": 28,
+                              "height": 28,
+                              "cornerRadius": 6,
+                              "justifyContent": "center",
                               "alignItems": "center",
                               "children": [
                                 {
-                                  "type": "text",
-                                  "id": "S849k",
-                                  "name": "set_lead_3",
-                                  "fill": "#9A9BA5",
-                                  "content": "Set as lead",
-                                  "fontFamily": "Inter",
-                                  "fontSize": 12,
-                                  "fontWeight": "normal"
-                                },
-                                {
-                                  "type": "text",
-                                  "id": "1E9tW",
-                                  "fill": "#00FF9C",
-                                  "content": "Prompt",
-                                  "fontFamily": "Inter",
-                                  "fontSize": 12,
-                                  "fontWeight": "normal"
-                                },
-                                {
-                                  "type": "text",
-                                  "id": "oa37n",
-                                  "fill": "#FF4D6D",
-                                  "content": "Remove",
-                                  "fontFamily": "Inter",
-                                  "fontSize": 12,
-                                  "fontWeight": "normal"
+                                  "type": "icon_font",
+                                  "id": "tCodD",
+                                  "width": 16,
+                                  "height": 16,
+                                  "iconFontName": "ellipsis",
+                                  "iconFontFamily": "lucide",
+                                  "fill": "#9A9BA5"
                                 }
                               ]
                             }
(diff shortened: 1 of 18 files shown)


yicheng47:
## C10 — Mission workspace UI: implementation plan

### Goal

A polished \`/missions/:id\` page that lets the human watch and steer a live mission — read the event feed, click ask-human cards, type into the lead's stdin via \`human_said\`, eyeball each runner's TUI, and reattach cleanly when the window reopens. The backend (C5–C8 + C9 CLI) already produces the events; C10 is the consumer.

Tracked in \`docs/impls/v0-mvp.md#C10\`. Manual-test plan in \`docs/tests/v0-mvp-tests.md\` (the workspace section).

### Scope of this PR

- Add \`src/pages/MissionWorkspace.tsx\` and route \`/missions/:id\` (mounted in \`src/App.tsx\` next to the existing routes).
- Subscribe to the three Tauri events the bus emits — \`event/appended\`, \`inbox/updated\`, \`watermark/advanced\` — filtered by \`mission_id\`.
- Render four panes:
  1. **Event feed** — message / signal / system / terminal-status rows with clear visual separation.
  2. **Ask-human cards** for pending \`human_question\` events (preserves \`payload.on_behalf_of\` attribution: \`@impl → @architect → you\`).
  3. **Mission input** that emits a \`signal human_said\` (broadcast or \`payload.target = <handle>\`).
  4. **Runner rail** showing each session, lead badge, status dot (busy/idle from C8 router status map), open-terminal action.
- **xterm-backed runner terminal**: same raw-byte \`session/output\` contract as the existing direct-chat pane (\`src/pages/RunnerChat.tsx\` is the reference impl). Lazy-mount per runner; one xterm per session id.
- **Reopen behavior**: loading \`/missions/:id\` fetches mission + session metadata, replays feed state from the bus's existing initial-replay mechanism (no extra backend), reattaches terminals to live sessions where present.

Out of scope: \`/missions\` list page and the Start Mission modal (C11). \`/debug\` cleanup also stays in C11.

### Frontend file layout

\`\`\`
src/pages/
  MissionWorkspace.tsx          # page shell, route handler, layout
src/components/mission/
  EventFeed.tsx                 # virtualized list of feed rows
  EventFeedRow.tsx              # row renderer with kind-aware variants
  AskHumanCard.tsx              # card with prompt/choices, attribution chain
  MissionInput.tsx              # textarea + target-handle picker, emits human_said
  RunnerRail.tsx                # rail with sessions, status dots, lead badge
  RunnerTerminal.tsx            # xterm pane bound to one session_id
src/lib/mission/
  feed.ts                       # event-by-id store, replay-then-tail merge
  pendingAsks.ts                # human_question + human_response → pending map
  runnerStatus.ts               # runner_status events → handle→status map
\`\`\`

The lib helpers stay separate from React components so the projection logic is testable without rendering. We've already paid this split for the direct-chat session list (\`src/lib/activeSessions.ts\`); same shape here.

### Event consumption

Three Tauri-event subscriptions, each filtered on \`mission_id\` since multiple missions can be open in the same process:

- \`event/appended\` → push the envelope into the per-mission feed store. The store is keyed on \`event.id\` (ULID) so initial replay + live tail merge with no dedupe logic — same id always wins. Order is append order, which matches insertion order from the bus.
- \`inbox/updated\` and \`watermark/advanced\` → route to the per-runner inbox indicators (only matter for the rail's "@impl has 3 unread" counters; the actual inbox is rendered by the agent itself via \`runner msg read\`).

Reopen path: the bus's \`for_mission\` already does an initial replay from offset 0 on mount, and the bus is mounted by \`mission_start\` and stays mounted until \`mission_stop\`. So when \`MissionWorkspace\` first mounts, it receives \`event/appended\` for *every* historical event in order — no separate replay endpoint needed. Same trick the bus already offers C8 router.

### Event-feed rendering

Three row variants:

- **Message** (\`kind = "message"\`): \`@from → @to: <text>\`. Broadcast (\`to == null\`) renders \`@from → all\`. Visually distinct from signals.
- **Signal** (\`kind = "signal"\`): \`<type>\` chip + condensed payload preview. Special handling for \`mission_start\`, \`mission_stopped\`, \`human_question\`, \`human_response\`, \`mission_warning\` (these get clearer labels). \`inbox_read\` is filtered from the feed by default — it's high-volume bookkeeping noise; behind a toggle if the user wants to see it.
- **Terminal-status row** (synthesized client-side from \`session/exit\` events): \`@coder exited (code=0)\` so a runner crash is visible in the feed, not just the runner rail.

Feed virtualization: the spec calls out backpressure as a risk (\`feed backpressure: agent output and event volume must not make React render unbounded rows on every chunk\`). Use \`react-window\` or a lightweight custom virtualizer (~50 lines) to render only the visible window. Decision in implementation: try without virtualization first, profile on a 5k-row mission, then add it if the framerate drops. If we virtualize, prefer a custom impl over react-window to keep the dep tree lean.

### Ask-human cards

When an \`event/appended\` arrives with \`signal_type == "human_question"\`:

- Look up the matching \`human_response\` (if any) in the same feed by \`human_response.payload.question_id == human_question.id\`. If matched: render as \`Answered: <choice>\` static row. If unmatched: render the interactive card.
- Card UI: prompt, attribution (if \`payload.on_behalf_of\` is set, render \`@<asker> → @lead → you\`; else \`@<asker> → you\`), one button per \`payload.choices[]\` value.
- Click action: call a new Tauri command \`mission_human_response\` (see backend below) which appends \`signal human_response\` with \`payload.question_id\` set. The C8 router observes the new event and routes the answer to the original asker's stdin.

Pending-ask matching is purely client-side and uses the canonical id contract from arch §5.5.0 (\`question_id == human_question.id\`). On reopen, the bus's initial replay re-delivers both \`human_question\` and any matching \`human_response\` so the card auto-answers itself; no router watermark is consulted by the UI.

### Mission input → human_said

A textarea at the bottom of the workspace, with an optional target-handle picker (defaults to broadcast = lead). Submit calls a new \`mission_human_said\` Tauri command that appends \`signal human_said\` with \`payload.text = <input>\` and optional \`payload.target = <handle>\`. The router's existing \`human_said\` handler then injects to either the chosen target or the lead.

### Runner rail

Rendered from the joined view of \`crew_runner::list\` (already shipped) + the live session map for this mission. Per row: handle, role, lead badge, status dot, and an "open terminal" button.

- Status dot: derived from \`runner_status\` signals seen in the feed (\`busy\` / \`idle\`). Default to \`busy\` for the lead at mission start, \`idle\` for workers; otherwise use the latest \`runner_status\` per handle. The router already keeps a per-mission status map; for v0 the UI can compute its own from the feed (cheap, replayable, doesn't need a new backend command).
- Lead badge: from \`crew_runners.lead\`. Comes from the \`StartMissionOutput\` already returned to the frontend — no extra round-trip.
- Open-terminal: lazy-mounts a \`RunnerTerminal\` for that session. Reuses the direct-chat pattern (xterm + \`session/output\` listener).

### Backend additions

Two small Tauri commands. Both are thin wrappers around the existing \`runner-core\` event-log writer; no new lifecycle, no router changes.

- \`mission_human_said(mission_id, text, target?)\` — appends \`signal human_said\` (\`from = "human"\`) to the mission's events log. \`payload.text\` mandatory, \`payload.target\` optional. The router's \`human_said\` handler picks it up.
- \`mission_human_response(mission_id, question_id, choice)\` — appends \`signal human_response\` (\`from = "human"\`) with \`payload.question_id\` and \`payload.choice\`. The router's \`human_response\` handler routes the answer to the original asker.

Both write through \`EventLog::append\` so they go through the same flock-guarded ULID assignment as the CLI. No DB writes; the bus + router observe the new line on the next tick.

### Reopen behavior in detail

- \`MissionWorkspace\` mount → \`mission_get(id)\` → metadata.
- \`crew_list_runners(crew_id)\` → roster for the rail.
- \`session_list({ mission_id })\` → live session ids per handle (already exposed).
- Subscribe to \`event/appended\` filtered by \`mission_id\`. The bus's initial replay re-delivers historical events; the feed renders them in append order. Pending asks reconstruct themselves because the human_question + human_response pairs reach the UI in order.
- For each running session, when the user opens its terminal: subscribe to \`session/output\` filtered by \`session_id\`, attach an xterm. The direct-chat code path (\`src/pages/RunnerChat.tsx\`) is the working reference — same critical-ordering note (listener attached BEFORE any spawn/attach call).

### Risks

- **Feed backpressure**, called out in the impl plan. Mitigation: feed store is incremental (id-keyed, no full-list re-renders); add windowing only if profiling shows the need.
- **Card dedupe across replay + live tail.** Solved by the canonical \`question_id == human_question.id\`: the same pair always reduces to "answered" if both events are present, regardless of arrival order.
- **xterm + raw bytes** vs feed: raw PTY output stays in xterm only — feed shows session lifecycle events (\`mission_start\`, \`session/exit\` synthesized rows), not stdout. Two distinct surfaces by design.
- **Backend command surface.** Adding two commands rather than reusing a generic "append signal" command — gives us a place to validate \`question_id\` is a real ULID and \`target\` is in the roster, mirroring the CLI's checks. Same validation constants the CLI uses.

### Order of work

1. Backend: \`mission_human_said\` + \`mission_human_response\` commands, with unit tests asserting append shape.
2. Lib helpers (\`feed.ts\`, \`pendingAsks.ts\`, \`runnerStatus.ts\`) with unit tests.
3. \`MissionWorkspace\` shell + route registration in \`App.tsx\`.
4. \`EventFeed\` + \`EventFeedRow\` (no virtualization yet).
5. \`AskHumanCard\` wired to the response command.
6. \`MissionInput\` wired to the human_said command.
7. \`RunnerRail\` + \`RunnerTerminal\` (xterm reuse).
8. Reopen path: refresh + assert feed and live terminals come back without a manual re-spawn.
9. Manual smoke test against a real two-runner crew (script the lead and impl to exercise the demo path from §172 of v0-mvp.md).
10. \`cargo fmt\` + clippy + tsc + lint.

### Out of scope (explicit)

- \`/missions\` list page + Start Mission modal (C11).
- Pending-ask indicator on the missions list — also C11.
- Threads / reactions / reply-to UI beyond \`@<handle>\` (v0.x).
- Multi-mission-per-crew (v0 ships one live mission per crew; the workspace assumes that invariant).
- LLM-driven message classification or signal inference (the router stays a hardcoded dispatcher; the UI projects raw events).

### Open questions

1. **Virtualization library vs hand-rolled.** Default plan: ship without virtualization, profile, decide. If needed, prefer a small hand-rolled windowed list (~50 LOC) over adding \`react-window\`. Flag if you'd prefer the dep up front.
2. **\`human_said\` from non-mission contexts.** Should the input be disabled when \`mission.status != "running"\`? My default: disable both input + ask-human buttons on a stopped mission, render the feed read-only. Confirm if you want it interactive (e.g. for retroactive notes — probably not, given the log is the audit record).
3. **Color/visual contract.** The Pencil designs in \`design/\` cover the runner / crew screens; I don't see an existing mission workspace mock. Plan: use the same Tailwind primitives + \`src/components/ui/\` palette, mirror the visual hierarchy of \`CrewEditor\`. Will flag in code where I've improvised.

yicheng47:
Review findings:

1. High: lead detection is wrong when the lead is not the first roster slot. `src/pages/MissionWorkspace.tsx:147` uses `sessions[0].handle`, while `src-tauri/src/commands/session.rs:78-92` orders by crew position and does not return the `lead` flag. If the user reassigns lead to slot 2, the rail badges the wrong runner and `MissionInput` defaults human messages to the wrong target.

2. Medium: switching between `/missions/:id` pages leaks the previous mission feed into the next mission. `src/pages/MissionWorkspace.tsx:65-68` always merges replayed events into prior `events`, and cleanup only resets `seenIdsRef` at line 104. The next mission feed, ask-human state, and runner-status map can include stale events.

3. Medium: stopping a mission leaves the UI in a running state. `src/pages/MissionWorkspace.tsx:253` discards the `mission_stop` result, so the status pill, Stop button, and input disabled state at line 313 stay based on the old `mission.status` until reload.

4. Medium: mission PTYs can miss startup output. The workspace fetches existing sessions before rendering terminals, while `src/components/RunnerTerminal.tsx:185-195` only listens for future `session/output` events. Any output or fast exit emitted between `mission_start` spawning sessions and the terminal listener attaching is gone, unlike the direct chat pre-spawn buffering path.

Verification:
- `pnpm run lint` passed.

yicheng47:
## Review of `1037035` + `877b808`

Verified: `pnpm tsc --noEmit` ✅, `pnpm lint` ✅, `cargo test -p runner --lib` 89 passed.

### Backend (`session.rs`, `manager.rs`, `commands/session.rs`)
- New `OutputEvent.seq` (monotonic per session) + `MAX_OUTPUT_BUFFER_CHUNKS = 4096` ring buffer per session in `SessionManager`. Every PTY chunk goes through `record_output` which assigns the seq + appends to the buffer.
- New `session_output_snapshot` Tauri command returns the bounded buffer for late attachers.
- `SessionRow` now includes `lead: bool` denormalized via `LEFT JOIN crew_runners` so the workspace doesn't have to look up the lead separately.
- Test in `manager.rs` covers both seq monotonicity and per-session filtering.

### Frontend
- `RunnerTerminal.tsx` — proper subscribe → snapshot → drain-pending flow keyed on `seq`. Listeners are registered before the snapshot RPC; live chunks during the round trip get buffered in `pendingLive` and merged after replay (dropping any seq ≤ what the snapshot already covered). On `active` flips, double-`requestAnimationFrame` so the layout settles before `fit/refresh/focus`. `onError` / `onExit` go through refs so prop-identity churn doesn't tear xterm down. **Tab switching now preserves real screen state** because xterm's own buffer carries through; the snapshot covers late attach.
- `RunnerChat.tsx` — fully refactored to mount one `RunnerTerminal` per direct session in a hidden-pane stack (just like the workspace). Switching chats no longer races the listener, and `userEndedRef` was replaced with a `Set<string>` so multiple concurrent kills each get the right "stopped" classification.
- `Sidebar.tsx` — when clicking a SESSION row that targets the route you're already on, `navigate(..., { replace: true, state: { sessionId } })` so RunnerChat's effect re-runs against the new requestKey instead of the navigation being a no-op.
- `MissionWorkspace.tsx` — switched the pane stack to `display: none` (Tailwind \`hidden\`) instead of z-index/visibility; now propagates `lead` from `SessionRow` to derive `leadHandle`; extracted `stopMission` callback so the inline confirm doesn't capture stale state.
- `lib/types.ts` — `SessionOutputEvent { seq, ... }` typed; `lib/api.ts` — `session.outputSnapshot()` and `SessionRow.lead` exposed.

### Potential follow-ups (not bugs blocking the smoke test)
1. `manager.rs`: drop `output_buffers[session_id]` and `output_seq[session_id]` inside `forget()` so terminated sessions don't hold their last 4096 chunks forever.
2. `MissionWorkspace.tsx` ~line 121: the `session/exit` listener fires `api.session.list` for *any* exit, including those of other missions. Cheap but worth filtering on this mission's `id`.

PR #22, opened 2026-04-27
validation: MVP smoke fixes

yicheng47:
Validation/fix branch after the v0 MVP merge.

Current change:
- Align Crew Detail slot action menu with Pencil node CUKjM.

Follow-up validation findings are tracked in PR comments.


yicheng47:
Validation finding to fix next:

Add Slot modal does not match Pencil node sYprG.

Current src/components/AddSlotModal.tsx is still a full create-runner form: handle, display name, role, runtime, command, args, cwd, and system prompt. The design expects a centered 560px modal with header/subtitle, Runner search input, selectable existing-runner list, per-slot prompt override toggle/copy, and simple footer actions.

This means the Add Slot flow should align with the shared-runner model: pick an existing runner for the crew slot, with slot-specific prompt override UI shown as off by default. Creating a brand-new runner can stay in the dedicated runner creation flow.

PR #23, opened 2026-04-28
validation: MVP follow-up fixes

yicheng47:
Follow-ups after PR #22. Changes: direct-chat SESSION tray only lists rows with direct_session_id; blank crew-name drafts show disabled Save instead of Saved; design/runners-design.pen updated to match the current validation design; repo-wide agent/developer conventions moved into AGENTS.md with CLAUDE.md as a compatibility pointer. Mission history is intentionally deferred to a separate designed surface; see the PR follow-up comment. Validation: pnpm exec tsc --noEmit; pnpm run lint; git diff --check.


yicheng47:
Follow-up for mission history:

The current runtime sidebar should stay aligned with the Pencil design and only show live runtime entries. For completed/past missions, add a designed history surface later rather than forcing a "mission history" row into the MISSION tray.

Candidate direction: add a history tab/row under the WORKSPACE section, or another explicit workspace-level history affordance, then route it to a proper mission-history page. This should be designed first and handled in a separate PR.

yicheng47:
Follow-up for agent-native session resume:

For each Runner session, persist the agent CLI native session/conversation id alongside our sessions row. Codex and Claude Code both have their own resumable session identity separate from Runner PTY/session id. We should capture that id when a direct chat or mission PTY starts, store it on the session entity, and use it on the next launch/reopen so closing the app does not force a fresh agent conversation.

Expected behavior:
- Direct chat: reopening the same runner can resume the prior Codex/Claude Code conversation when available.
- Mission PTY: reopening a mission after app restart can resume each runner agent session using the stored native id.
- If the native id is missing or resume fails, fall back to the current fresh-spawn behavior and surface a clear warning.

This likely needs runtime-specific adapters for how to discover/pass the native session id for codex and claude-code, plus a schema migration for the persisted session field.

yicheng47:
Review findings against local HEAD `530bdd7` (`feature(session): direct-chat redesign with resume + multi-chat per runner`). Note: this commit is not pushed to `origin/mvp-validation-followups` yet, so these line refs correspond to the local branch state.

1. High: direct-chat spawn can still hang in dev StrictMode. `src/pages/RunnerChat.tsx:343` gates the effect with `startedKeyRef` before `session_start_direct` resolves. In React StrictMode, enabled at `src/main.tsx:7`, the second effect setup returns because the key is already set; then the first async spawn resolves under `cancelled`, kills the session, and nothing starts the visible replacement. Result: new direct chat can stay on `Starting...`.

2. High: `session_resume` is not atomically claimed. `src-tauri/src/session/manager.rs:646` reads `status`, spawns a child, then `manager.rs:800` updates the row unconditionally. Two fast Resume clicks or two windows can both see `stopped`, spawn two PTYs for the same session id, and race the live map/output stream. This needs a per-session lock or a DB compare-and-set claim before spawning.

3. Medium: Codex session-key capture misses common cases. It only starts capture when `resolved_cwd` is `Some` at `src-tauri/src/session/manager.rs:591`, so Codex chats that inherit the process cwd never become resumable. Also `src-tauri/src/session/codex_capture.rs:91` marks a rollout file as tried before it has successfully parsed `session_meta`; if Codex creates the file before writing the first line, the watcher never retries it.

4. Medium: the stopped-session UI always says the prior conversation is preserved. `src/pages/RunnerChat.tsx:732` does not use `DirectSessionEntry.resumable`, so shell sessions and Codex rows with missing capture still promise resume semantics even though backend will start fresh.

Validation passed:
- `pnpm exec tsc --noEmit`
- `pnpm run lint`
- `cargo test --workspace`
- `git diff --check origin/mvp-validation-followups..HEAD`

yicheng47:
Follow-up review against local HEAD `36a2b83` (`feature(session): direct-chat redesign with resume + multi-chat per runner`). Note: this commit is still not pushed to `origin/mvp-validation-followups`, so these line refs correspond to the local branch state.

1. Medium: live chat opened from the Runners page is seeded as stopped. `src/pages/Runners.tsx:85` navigates with only `{ sessionId }`, but `src/pages/RunnerChat.tsx:370` defaults missing `sessionStatus` to `"stopped"`. Since `direct_session_id` only points at a running session, this path should pass `sessionStatus: "running"` or RunnerChat should infer running for this legacy attach path. Otherwise the terminal can briefly mount disabled/stopped, and if metadata refresh fails it stays wrong.

2. Low: `docs/impls/direct-chats.md` is stale relative to the implementation. It still says stopped sidebar rows auto-call `api.session.resume` before navigation at `docs/impls/direct-chats.md:151`, but the code now navigates first and shows an explicit Resume/Restart overlay. It also still lists Codex key capture as deferred at `docs/impls/direct-chats.md:220`, even though this patch adds `codex_capture.rs`.

The previous StrictMode, resume-claim, Codex partial-write, and non-resumable copy issues look addressed.

Validation passed:
- `pnpm exec tsc --noEmit`
- `pnpm run lint`
- `cargo test --workspace`
- `git diff --check origin/mvp-validation-followups..HEAD`

yicheng47:
Follow-up review against local HEAD `c41faac` (`feature(session): direct-chat redesign with resume + multi-chat per runner`). Note: this commit is still not pushed to `origin/mvp-validation-followups`, so these line refs correspond to the local branch state.

1. Medium: Codex key capture can attach the wrong native session when two Codex chats start in the same cwd close together. `src-tauri/src/session/codex_capture.rs:113` accepts the first rollout whose cwd matches, and `src-tauri/src/session/codex_capture.rs:217` only rejects rollouts older than `started_at - 5s`. A second chat started within that window can capture the first chat’s `payload.id`, so two Runner session rows may resume into the same Codex conversation.

2. Low: `docs/impls/direct-chats.md` still has stale migration-step text. `docs/impls/direct-chats.md:199` says “click-to-resume” and `docs/impls/direct-chats.md:201` says the sidebar pre-resumes, but the actual implementation now navigates first and resumes explicitly from the chat overlay.

The prior Runners-page attach issue is fixed, and the Codex capture doc section was mostly updated.

Validation passed:
- `pnpm exec tsc --noEmit`
- `pnpm run lint`
- `cargo test --workspace`
- `git diff --check origin/mvp-validation-followups..HEAD`

PR #25, opened 2026-04-30
validation: mission resume robustness + paused overlay

yicheng47:
## Summary

- **Workspace input gating** — when a mission row is `running` but every PTY is dead, the feed input is replaced by a bottom-anchored Resume card (`SessionEndedOverlay` `variant=\"inline\"`, mission-level copy). Feed and slot panes share one recovery anchor.
- **Reset cleanup** — `session_list` now filters on `archived_at IS NULL`, so post-reset sidebars no longer stack the old stopped row alongside the freshly-spawned one for every slot.
- **Lead launch prompt deferred** — `Router::inject_and_submit_delayed` lifts the bus's synchronous `mission_goal` injection onto the same 2.5s budget `schedule_first_prompt` uses for workers, so the lead's prompt no longer gets eaten by claude-code's boot / trust-folder screen on warm starts (mission_reset, fast mission_start).
- **Resume fresh-fallback for claude-code** — `claude_code_conversation_exists(cwd, uuid)` checks `~/.claude/projects/<encoded-cwd>/<uuid>.jsonl` (encoding maps both `/` and `.` to `-`). On miss, `SessionManager::resume` swaps `--resume` for `--session-id <existing-uuid>`, keeping the row's UUID bound to the new conversation. `Router::fire_lead_launch_prompt` reads the latest `mission_goal` text from the log and re-fires `compose_launch_prompt` for the lead — `session_resume` calls it when `SpawnedSession.fresh_fallback_lead` is set, so the architect gets the rich launch prompt (system_prompt + mission goal + roster + crew context), not the worker preamble.

## Test plan

- [ ] Start a mission with claude-code lead + impl. Verify the architect tab receives its launch prompt on a fresh start.
- [ ] Stop the mission, then Resume from the topbar. Verify both PTYs come back via `--resume <uuid>` (no fresh spawn) and the prior conversation continues.
- [ ] Reset a mission via the kebab. Verify the sidebar shows exactly one session per slot (no ghosts) and the lead receives the launch prompt 2.5s after spawn.
- [ ] Stop a mission, manually delete the lead's `~/.claude/projects/<encoded-cwd>/<uuid>.jsonl`, then Resume. Verify the lead falls back to a fresh spawn keeping its UUID and receives the rich launch prompt (not the worker coordination preamble).
- [ ] With the mission running, click Stop. Verify the feed shows the \"Mission paused\" overlay, the input is non-interactive, and clicking Resume from the overlay restores both PTYs.

### DIFF
--- a/cli/src/roster.rs
+++ b/cli/src/roster.rs
@@ -29,11 +29,26 @@ pub fn load(event_log: &Path) -> Option<Vec<RosterEntry>> {
     serde_json::from_slice::<Vec<RosterEntry>>(&bytes).ok()
 }
 
+/// Reserved virtual handle. The human isn't a slot in any crew, but
+/// they ARE a participant in the mission's bus traffic — workers can
+/// post `runner msg post --to human "<reply>"` and the workspace UI
+/// renders it as a `kind:"message"` row in the feed. Without
+/// reserving it here, `--to human` would either be rejected (when a
+/// roster sidecar is present) or silently allowed (sidecar missing)
+/// inconsistently. This makes it a first-class destination.
+pub const HUMAN_HANDLE: &str = "human";
+
 /// Returns true if `handle` is in the roster. Missing-sidecar is
 /// permissive for the same reason as the allowlist (a missing roster
 /// sidecar means something is broken upstream — don't strand the
 /// mission). Stderr warning surfaces the gap.
+///
+/// `HUMAN_HANDLE` always passes regardless of sidecar contents — see
+/// the doc on the constant for the rationale.
 pub fn is_known(event_log: &Path, handle: &str) -> bool {
+    if handle == HUMAN_HANDLE {
+        return true;
+    }
     match load(event_log) {
         Some(list) => list.iter().any(|r| r.handle == handle),
         None => {
--- a/design/runners-design.pen
+++ b/design/runners-design.pen
@@ -550,6 +550,51 @@
                   "fontWeight": "normal"
                 }
               ]
+            },
+            {
+              "type": "frame",
+              "id": "yE3cM",
+              "name": "sidebar_spacer",
+              "width": "fill_container",
+              "height": "fill_container"
+            },
+            {
+              "type": "frame",
+              "id": "VyTE6",
+              "name": "sidebar_div",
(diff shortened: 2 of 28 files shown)


yicheng47:
Review findings:

1. High: `cargo test --workspace` is red on this PR. Failing tests: `router::tests::mission_goal_injects_composed_prompt_to_lead`, `router::tests::fresh_mission_start_does_not_call_reconstruct_so_mission_goal_fires`, and `session::manager::tests::resume_reuses_row_and_preserves_agent_session_key`. The first two now see zero pushes because `mission_goal` moved to delayed async injection at `src-tauri/src/router/handlers.rs:67`. The resume test changes the saved Claude session key because the new conversation-file existence check falls back to fresh spawn when the test has no `~/.claude` file.

2. High: `mission_reset` is not all-or-nothing. After it kills existing PTYs, archives rows, wipes the event log, and flips the mission back to `running`, the spawn loop at `src-tauri/src/commands/mission.rs:987` uses `?` directly. If any later slot fails to spawn, earlier fresh PTYs remain alive, the bus/router are not mounted, old rows are archived, and the mission is left half-reset. This needs rollback similar to `mission_start`.

3. Medium: partial PTY failure is rendered as full mission pause. `allSessionsLive` is used for the status pill, input disable, and pause overlay at `src/pages/MissionWorkspace.tsx:340`. If only one slot crashes while the lead is still running, the UI says `stopped`, disables the feed input, and shows “All slots are stopped.” That blocks valid human-to-live-runner messages.

4. Medium: Codex resume still replays `runner.system_prompt` as a new positional prompt. In `SessionManager::resume`, `system_prompt_args(...)` is appended unconditionally after `codex resume <uuid>` at `src-tauri/src/session/manager.rs:947`. For Codex, that means every resume adds the runner brief as a new user turn.

5. Medium: live `runner/activity` still queries the removed `crew_runners` table. The emitter falls back to `crew_count = 0` at `src-tauri/src/session/manager.rs:1661`, so spawn/resume/exit events can temporarily publish wrong runner usage counts after the slots migration.

Verification:
- `pnpm run lint` passed.
- `cargo test --workspace` failed with the three tests listed above.

PR #28, opened 2026-05-01
feature: auto-update via tauri-plugin-updater

yicheng47:
## Summary
- One `UpdateContext` shared by a top-center `UpdateToast` and the Settings → Updates pane (Quill pattern). Toast just routes the user into Settings; the pane owns the download / install / restart flow.
- `tauri-plugin-updater` registered with `endpoints` pointing at the GitHub Releases `latest.json`, plus `bundle.createUpdaterArtifacts: true` so future releases ship `.sig` files. Capability gets `updater:default` (`process:allow-restart` was already there).
- `useUpdateChecker` hook drives the state machine (idle → checking → available → downloading → ready / error). `UpdateProvider` auto-checks ~3s after mount; if the auto-install toggle is on, it auto-advances to download.
- `UpdatesPane` rewired to a single status-driven row that swaps action (Check now / Download & install / Restart-to-update) plus an inline progress bar during download. Stale channel + notify-before-install rows dropped from the design too.
- Windows dropped from the release workflow — the PTY / WSL integration isn't ready and shipping a non-PTY build was misleading.

## Notes
- **Pubkey placeholder.** `src-tauri/tauri.conf.json` ships with `pubkey: "REPLACE_WITH_TAURI_SIGNING_PUBLIC_KEY"`. Paste in the real minisign pubkey from the keypair already wired into the CI secret before tagging v0.1.1, otherwise the updater rejects every payload.
- **First reachable version is v0.1.1.** v0.1.0 binaries don't have the plugin baked in, and that release also lacks `.sig` files (we forgot `createUpdaterArtifacts` last time). v0.1.0 → v0.1.1 is a manual download; from v0.1.1 onward it's in-app.

## Test plan
- [ ] `pnpm exec tsc --noEmit` (clean)
- [ ] `pnpm exec eslint src/` (clean)
- [ ] `cargo check` + `cargo clippy -- -D warnings` (clean)
- [ ] `pnpm build` (clean)
- [ ] Local dev: open Settings → Updates pane, confirm idle "You're up to date" + working "Check now" button
- [ ] Tag v0.1.1 once pubkey is swapped; verify `latest.json` + `.sig` artifacts attached to the release
- [ ] On a v0.1.1 install: bump to v0.1.2 in a follow-up tag, confirm toast appears and "Update" button opens Settings, then Download & install + Restart actually upgrades the running binary


yicheng47:
Review findings:

1. High: disabling auto-install does not actually disable it. `src/components/SettingsModal.tsx:178` persists the toggle as `"1"` / `"0"`, but `src/contexts/UpdateContext.tsx:35` and `src/contexts/UpdateContext.tsx:53` treat anything except `"false"` as enabled. So when the user turns “Install updates automatically” off, the next app launch still auto-checks, and if an update is found it still auto-downloads/installs. Use the same boolean encoding on both sides, or centralize the parser.

Verification:
- `pnpm run lint` passed with one Fast Refresh warning in `UpdateContext.tsx`.
- `cargo check` passed.
- `pnpm build` passed with the existing large chunk warning.

PR #29, opened 2026-05-03
dev: isolate dev-build data dir from prod install

yicheng47:
## Summary
- Dev builds (`cfg!(debug_assertions)`) now write to `~/Library/Application Support/com.wycstudios.runner-dev/` so `make dev` doesn't share state with a packaged v0.1.0 install
- Mirrors Quill's pattern: same identifier in `tauri.conf.json`, swap happens once in Rust `setup` and propagates via `AppState.app_data_dir`
- Affected paths: `runner.db`, bundled CLI at `<dir>/bin/`, mission event logs, session PATH prepend

## Test plan
- [x] `make dev` creates `com.wycstudios.runner-dev/` with a fresh `runner.db`
- [x] Existing `com.wycstudios.runner/` install untouched after dev run
- [ ] Release bundle (next tag) continues to use `com.wycstudios.runner/`

### DIFF
--- a/design/runners-design.pen
+++ b/design/runners-design.pen
@@ -24012,10 +24012,10 @@
                         {
                           "type": "text",
                           "id": "M9HaRG",
-                          "fill": "#9A9BA5",
+                          "fill": "#00FF9C",
                           "textGrowth": "fixed-width",
                           "width": "fill_container",
-                          "content": "You're up to date — last checked 12 minutes ago.",
+                          "content": "v0.1.1 is available — released 2 hours ago.",
                           "fontFamily": "Inter",
                           "fontSize": 11,
                           "fontWeight": "normal"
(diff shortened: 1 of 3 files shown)


yicheng47:
Review findings:

1. Low: dev-data cleanup docs use the wrong directory name. `docs/tests/v0-mvp-tests.md:64` tells smoke testers to remove `$HOME/Library/Application Support/com.wycstudios.runner.dev`, but this PR writes debug app data to the sibling path produced in `src-tauri/src/lib.rs:54`, e.g. `$HOME/Library/Application Support/com.wycstudios.runner-dev`. Anyone following the reset leaves the actual dev DB/logs/CLI in place, so tests can start from stale state. Update the doc to the hyphenated path.

Verification:
- `git diff --check origin/main...origin/dev/separate-dev-data` passed.

yicheng47:
Fixed in e47fab4: renamed `com.wycstudios.runner.dev` → `com.wycstudios.runner-dev` in the smoke-test reset block, and dropped the `com.wycstudios.runner` (prod-dir) wipe — after this PR `pnpm tauri dev` never writes there, so leaving it in the dev reset block was a footgun that would silently nuke a packaged install. Added a one-line note explaining the split.

PR #30, opened 2026-05-03
feature: ship default Build squad crew on first launch

yicheng47:
## Summary
- Adds migration `0002_default_crew.sql` that seeds a starter **Build squad** crew (architect lead + implementer + reviewer) on first launch
- Architect runs `claude-opus-4-7` / `high`; impl + reviewer run `claude-sonnet-4-6` / `medium` — also demos the per-runner model/effort feature
- Idempotent: seed is gated on `WHERE NOT EXISTS (SELECT 1 FROM crews)` so existing installs are untouched, and per-row `NOT EXISTS` guards make reseeds safe
- Seeds run from `open_pool` only — `open_in_memory` (tests) skips them so command tests still start from an empty DB
- Polishes `tests/fixtures/`: rename `feature-delivery.seed.sh` → `build-squad.seed.sh` with reviewer slot, model/effort, and a `--dev` flag pointing at the `com.wycstudios.runner-dev/` data dir
- Normalizes the three system prompts (architect / impl / reviewer) into a shared "Talking to the human" stanza

## Test plan
- [x] `cargo test -p runner` — 107 passing (3 new: empty-DB seed, skip-when-user-has-crew, idempotent reseed)
- [x] Local dev launch creates `com.wycstudios.runner-dev/runner.db` with the Build squad pre-seeded
- [ ] Verify packaged build also seeds on first launch (next release tag)


yicheng47:
Review findings:

1. Medium: the default crew is not a one-time seed. `src-tauri/src/db.rs:54` runs `seed_defaults` on every `open_pool`, and `src-tauri/migrations/0002_default_crew.sql:19` only checks whether `crews` is currently empty. If a user deletes Build squad and has no other crews, the next app launch recreates it. This contradicts “first launch” behavior; record an applied seed marker or otherwise distinguish “never seeded” from “user deleted all crews.”

2. Medium: runner-handle collisions can create a broken partial Build squad. The crew insert only gates on no crews, but the runner inserts skip if handles like `architect` already exist (`src-tauri/migrations/0002_default_crew.sql:89`). The slot inserts then require the fixed default runner IDs (`src-tauri/migrations/0002_default_crew.sql:231`), so an existing `@architect` runner with no crews yields a Build squad with no lead. The start modal treats any nonzero slot count as launchable (`src/components/StartMissionModal.tsx:117`), then the backend rejects it for no lead (`src-tauri/src/commands/mission.rs:164`). Skip the whole seed on conflicting runner handles, attach slots to existing runners, or gate on no existing user data.

Verification:
- `bash -n tests/fixtures/crews/build-squad.seed.sh` passed.

yicheng47:
Addressed both findings in e2d5cb6:

**1. Persistent seed marker.** Added an `_app_state` table with a `default_crew_seeded` row. `seed_defaults` now exits early when the marker is set, regardless of whether the user has crews or not. Even if they delete Build squad and have zero crews afterward, we never recreate it. Marker is set inside the same transaction as the seed itself, so a crash mid-seed rolls back both.

**2. Gate on any prior data, not just empty crews.** The seed now applies only when `crews == 0 AND runners == 0`. If either is non-empty (e.g. the user ran build-squad.seed.sh against this DB before opening the app, or manually created an `architect` runner), we skip the seed entirely and still set the marker — so we never produce a partial/leadless Build squad and we never reconsider the decision.

The migration SQL itself is now plain INSERTs (the per-row `NOT EXISTS` clauses became dead weight once Rust guarantees a clean target).

Test coverage:
- `seed_defaults_skips_when_user_has_a_runner_but_no_crew` — covers the partial-crew failure mode that motivated finding #2
- `seed_defaults_marker_prevents_reseed_after_user_deletes_everything` — covers finding #1
- existing `seed_defaults_inserts_build_squad_on_empty_db` and `seed_defaults_is_idempotent_across_reseeds` still pass

`cargo test -p runner` → 109 passing. `cargo clippy -p runner -- -D warnings` → clean.

PR #36, opened 2026-05-03
fix: resolve login-shell PATH so GUI-launched app finds agent CLIs (#35)

### LINKED ISSUE #35: bug: GUI-launched app can't spawn agent commands ("claude" not on launchd PATH)
## Description

A packaged Runner build launched from Finder / Spotlight / the Dock
fails to start any mission whose runner uses an agent CLI that lives
outside the launchd default PATH (Homebrew, npm-global, mise, nvm,
fnm, asdf, etc.). The mission start fails immediately with:

```
spawn claude: Unable to spawn claude because it doesn't exist on the filesystem and was not found in PATH
```

This hits every user who downloads the v0.1.1 .dmg from the release
page and tries the seeded Build squad — `claude` is the default
command for the seeded architect/impl/reviewer runners. From the
user's perspective the app is broken out-of-the-box.

## Expected behavior

A child PTY spawned by the app should be able to find `claude` (or
any other agent CLI) on PATH whenever the user's login shell can
find it, regardless of how the app itself was launched.

## Steps to reproduce

1. Install Runner v0.1.1 from the .dmg.
2. Launch from Finder / Spotlight / the Dock (NOT from a terminal
   via `open -a` after sourcing your shell rc).
3. Click "Start mission" against the seeded Build squad.
4. Mission boot fails immediately; the workspace shows the spawn
   error above. Same crash happens for `codex`, any agent CLI not on
   `/usr/bin:/bin:/usr/sbin:/sbin`.

## Root cause

macOS launchd hands GUI apps a stripped PATH — typically just
`/usr/bin:/bin:/usr/sbin:/sbin`. The user's shell PATH (Homebrew at
`/opt/homebrew/bin` or `/usr/local/bin`, npm-global at
`~/.npm-global/bin` or `~/.nvm/versions/...`, mise / asdf / fnm
shims, etc.) is loaded by the shell rc, not by launchd, so the
Tauri process never sees it.

In `pnpm tauri dev` from a terminal this is invisible because the
terminal-spawned process already has the shell PATH. Production
.dmg launches don't.

We already prepend `<app_data_dir>/bin` to the child's PATH so the
bundled `runner` CLI resolves
(`src-tauri/src/session/manager.rs:355-368` for missions, `:603-613`
for direct sessions), but `claude` and friends are not bundled — we
expect them on the user's PATH.

## Relevant code

- `src-tauri/src/session/manager.rs:340-388` — mission spawn path,
  where the error message is formatted (line 388:
  `spawn_command(cmd).map_err(|e| Error::msg(format!("spawn {}: {e}",
  runner.command)))`).
- `src-tauri/src/session/manager.rs:560-613` — direct-session spawn
  path with the same PATH wiring.
- `src-tauri/src/lib.rs` — app `setup` block, where a shell-PATH
  resolution step would slot in alongside `cli_install::install_runner_cli`.

## Workarounds users can apply today

- Edit each runner template's `command` field to the absolute path
  of the agent CLI (e.g. `/opt/homebrew/bin/claude` instead of
  `claude`). Per-runner override; survives app restart.
- Symlink the agent CLI into `/usr/local/bin/` so it's visible
  from any GUI-launched app.
- Launch the app from a terminal so it inherits the shell PATH:
  `open -a Runner.app` from a shell that has `claude` on PATH.

None of these are reasonable defaults for a fresh download. We
should fix the root cause.

## Fix sketch

At app startup (in `lib.rs::setup`, before the AppState is built),
resolve the user's login-shell PATH and cache it on AppState:

1. Read `$SHELL` (fall back to `/bin/zsh` on macOS).
2. Spawn `$SHELL -ilc 'printf "%s" "$PATH"'` with a short timeout
   (~2s) and a clean env. The `-i` (interactive) and `-l` (login)
   flags are needed to source `~/.zshrc` and `~/.zprofile`
   respectively, which is where Homebrew shellenv / nvm / mise
   inject their PATH entries.
3. Capture stdout, trim, and store as `AppState.shell_path`.
4. In `session/manager.rs`, when building the child env, merge:
   `<shim>:<bin_dir>:<shell_path>:<inherited PATH>`. The shell PATH
   sits after the bundled bin (so our `runner` shim wins) but before
   the inherited launchd PATH.
5. Cache resilience: if the resolution fails or times out, fall
   back to the current behavior with a one-line `eprintln!` log so
   the user sees the issue in `Console.app`. The app shouldn't
   refuse to launch over a PATH-resolution stall.

This mirrors how VS Code, GitHub Desktop, and most other
Electron/Tauri apps handle the same problem (the `shell-env` npm
package is the JS equivalent).

Out of scope for the fix: bundling the agent CLIs (too heavy);
auto-detecting common install locations (brittle and not as
correct as `$SHELL -ilc` since it can't know about user-specific
managers like mise/asdf).

## Environment

- OS: macOS (any version where launchd strips PATH for GUI apps,
  i.e. all current versions)
- Version: 0.1.1

yicheng47:
## Summary

- Resolves [#35](https://github.com/yicheng47/runner/issues/35): a packaged Runner launched from Finder/Dock inherits launchd's stripped PATH (`/usr/bin:/bin:/usr/sbin:/sbin`), so seeded Build squad runners can't find `claude` / `codex` / mise shims / etc. and every mission boot fails with `spawn claude: ... not found in PATH`.
- New `shell_path` module resolves the user's login-shell PATH at app startup via `$SHELL -ilc 'printf "%s" "$PATH"'` with a 2s timeout, falling back gracefully on failure / Windows. Logs a one-liner to stderr on the fallback paths so it surfaces in `Console.app`.
- `SessionManager` stores the captured PATH and merges it into every child PTY's PATH (mission spawn, direct chat, resume) via a new `compose_path(shim, bin)` helper. Order: **shim ► bin ► shell ► parent** — the bundled-runner shim still wins, the inherited launchd PATH stays as the tail.

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test --workspace` — 154 passing (incl. new `compose_path_orders_segments_and_skips_empties` covering shim/bin/shell ordering, empty shell PATH, empty parent PATH)
- [x] `pnpm tsc --noEmit` and `pnpm run lint` clean (one pre-existing unrelated warning in `UpdateContext.tsx`)
- [ ] Manual: package a build, install via .dmg, launch from Finder, confirm Build squad mission starts without the `spawn claude` error
- [ ] Manual: `pnpm tauri dev` from a terminal still works (this path was never broken; verify no regression)

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -6,6 +6,7 @@ mod event_bus;
 mod model;
 mod router;
 mod session;
+mod shell_path;
 
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -88,10 +89,19 @@ pub fn run() {
                 eprintln!("runner: failed to install bundled CLI: {e}");
             }
 
+            // Resolve the user's login-shell PATH once at startup so
+            // child PTYs can find tools that live outside launchd's
+            // stripped default PATH (Homebrew, mise/asdf/fnm, npm-global,
+            // etc.). Best-effort: a failure or timeout just leaves
+            // shell_path = None and we fall back to the inherited
+            // launchd PATH. See `shell_path` module docs for the
+            // launchd-strips-PATH problem this fixes.
+            let shell_path = shell_path::resolve_login_shell_path();
+
             app.manage(AppState {
                 db: pool,
                 app_data_dir,
-                sessions: session::SessionManager::new(),
+                sessions: session::SessionManager::new(shell_path),
                 buses: event_bus::BusRegistry::new(),
                 routers: router::RouterRegistry::new(),
             });
--- a/src-tauri/src/session/manager.rs
+++ b/src-tauri/src/session/manager.rs
@@ -201,6 +201,13 @@ pub struct SessionManager {
     /// SIGTERM produces a non-zero exit code. Entries are cleared by
     /// the reader after the DB row is updated.
     killed: Mutex<HashSet<String>>,
+    /// User's login-shell PATH, captured once at app start by
+    /// `shell_path::resolve_login_shell_path`. None when the resolve
+    /// failed/timed out, when running on Windows, or in tests.
+    /// Merged into every child PTY's PATH so GUI-launched apps can
+    /// find tools (claude, codex, mise, etc.) that aren't on
+    /// launchd's stripped default PATH.
+    shell_path: Option<String>,
 }
 
 /// RAII guard that releases a `resuming_claims` entry on drop. The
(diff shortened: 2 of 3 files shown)


yicheng47:
Review findings:

1. Medium: login-shell PATH capture can return a corrupted/non-PATH string. `src-tauri/src/shell_path.rs:36` runs the user’s interactive shell and trusts all stdout from `printf '%s' "$PATH"` at `src-tauri/src/shell_path.rs:47`. This breaks for shells like `fish` where `$PATH` expansion is not POSIX colon-delimited, and for any `.zshrc` / `.bashrc` that prints banners or tool messages to stdout. In those cases the packaged GUI app can still fail to find `claude` / `codex`. Prefer a shell-neutral `env`/`printenv` capture and parse a clean `PATH=` line or marker-delimited value.\n\n2. Low: the timeout path does not kill the shell process. `src-tauri/src/shell_path.rs:30` moves `Command::output()` into a background thread, and `src-tauri/src/shell_path.rs:67` returns after timeout without terminating the child. A hung shell init can leave a background login shell running after every app launch.\n\nVerification passed:\n- `git diff --check origin/main...origin/fix/gui-launch-shell-path`\n- `cargo fmt --all --check`\n- `cargo clippy --workspace --all-targets -- -D warnings`\n- `cargo test --workspace`\n- `pnpm exec tsc --noEmit`\n- `pnpm run lint` passed with the existing Fast Refresh warning.

PR #60, opened 2026-05-05
fix: misc workspace + chat polish

yicheng47:
## Summary

Four small UX fixes that piled up while smoke-testing post-v0.1.1.

### `a0fd125` — suppress "Mission paused" overlay during archive
The paused overlay gated only on `mission.status === "running" && !allSessionsLive && !resumingAll`. None of those flip when an archive starts, so the user briefly saw "Mission paused / Resume" between clicking Archive and the route swap to `/runners`. Added `!archivingMission` so the centered ArchivingOverlay owns the visual during the in-flight RPC.

### `cf1d7fb` — make Default working directory editable
The Settings General pane shipped Default working directory as a non-interactive `DisabledDropdown` with a "stub — no backend yet" hint, even though Default crew (also localStorage-only) is interactive. Added a real folder picker via `@tauri-apps/plugin-dialog`'s `open({ directory: true })`. Persists to `localStorage["settings.defaultWorkingDir"]` mirroring the Default crew pattern. The picked path is an absolute path the OS confirmed exists. An X-button clears the stored value back to "no default."

### `d61a2bb` — archive overlay scrim + initial (wrong-direction) chat header parity attempt
Two parts. First half is the actual fix: the mission workspace's archive overlay rendered as a small amber pill with no scrim, so the destructive transition flashed briefly over a still-live-looking feed. Pass `withScrim` (matching `RunnerChat`'s archive flow) so the feed dims behind the pill.

Second half was wrong-direction — I changed the chat header typography to match the mission header. The follow-up commit (`8fcdabf`) reverts that part to the chat's own canonical design. Net effect: only the archive scrim survives.

### `8fcdabf` — chat header matches Pencil design `u6woG`/`vS5ce`
- **Typography**: title is `<span className="font-mono text-[15px] font-semibold">` per node `U9Fx8f`. Title-stack gap `[3px]`, title-row gap `2.5`, meta gap `2`, meta color `text-fg-2`, avatar accent on the icon (matching `dnPId`).
- **Kebab menu**: new `ChatKebab` + `KebabItem` mirroring `MissionKebab` shape. Pin / Rename / Archive items mapped 1:1 to the design's `session_ctx_menu` (`P5CLA` running / `L31Zb` stopped). Reset is omitted (mission-only — a chat has no slots to respawn). New `togglePin` + `renameChatPrompt` callbacks; `archiveChat` reused as-is.
- **Stopped-state action**: "Resume" instead of "Back to runner" when the row is resumable. Matches Pencil node `HLXK6` (`#0F2418` fill / `#1F4D33` border / accent text + Play glyph). Same action the inline SessionEndedOverlay card already fires; surfacing it in the topbar lets the user recover without scrolling. Falls back to "Back to runner" only when the row is genuinely non-resumable or no session exists yet.

## Test plan

- [x] `cargo test --lib`: 160 passed (no backend changes — should still pass).
- [x] `pnpm tsc --noEmit`: clean.
- [x] `pnpm lint`: clean except the pre-existing `UpdateContext.tsx` Fast-Refresh warning.

### Manual

- [ ] **Archive a paused mission**: stop a Build squad mission so the "Mission paused" overlay shows, then click the topbar kebab → Archive. Should NOT flash "Mission paused" — should show the amber Archiving pill over a dimmed feed, then navigate to `/runners`.
- [ ] **Settings → Default working directory**: open Settings, click the `~/` button, native folder picker opens, pick a folder, the button shows the chosen path. X-button clears it back to `~/`.
- [ ] **Chat header — running**: open a chat, the topbar shows `@handle` in JetBrains Mono 15px, CHAT pill, accent status pill, and the kebab. Click the kebab → Pin / Rename / Archive items render. Pin toggles the sidebar pin state; Rename prompts and updates the title in place.
- [ ] **Chat header — stopped**: stop a chat. The topbar action becomes a green "Resume" button (matching `HLXK6`). Click it → terminal resumes.
- [ ] **Chat header — non-resumable**: if a chat row genuinely lacks an `agent_session_key`, the action is still "Back to runner" (no-op fallback).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,15 +35,17 @@ jobs:
           components: clippy, rustfmt
       - uses: Swatinem/rust-cache@v2
       # Tauri 2 on Linux needs webkit2gtk and friends even for `cargo check`
-      # because tauri-runtime-wry links against them.
-      - name: Install system dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y \
-            libwebkit2gtk-4.1-dev \
-            libgtk-3-dev \
-            libayatana-appindicator3-dev \
-            librsvg2-dev
+      # because tauri-runtime-wry links against them. The `cache-apt-pkgs`
+      # action caches the resolved .deb tree in GitHub's actions cache,
+      # which cuts the install step from ~45s (apt mirror download +
+      # dpkg) to ~3s (untar from cache) on every subsequent run. Bump
+      # the `version` key to invalidate the cache when the package
+      # list below changes.
+      - name: Install system dependencies (cached)
+        uses: awalsh128/cache-apt-pkgs-action@v1
+        with:
+          packages: libwebkit2gtk-4.1-dev libgtk-3-dev libayatana-appindicator3-dev librsvg2-dev
+          version: 1.0
       # Tauri's build script validates that every `bundle.externalBin`
       # entry exists for the active target on every cargo build,
       # including clippy / check / test. CI doesn't bundle, but it does
--- a/src-tauri/src/commands/session.rs
+++ b/src-tauri/src/commands/session.rs
@@ -296,9 +296,15 @@ pub async fn session_archive(
 /// `None` (or an all-whitespace string, treated as None) to revert to
 /// the auto-derived label (`@handle · <time>`). Trims surrounding
 /// whitespace before persisting.
+///
+/// Emits a `session/updated` Tauri event after the row flips so the
+/// sidebar's CHAT list can refresh — without it, renaming from the
+/// chat-page kebab would update the row but leave the sidebar's
+/// title stale until some other refresh trigger fires.
 #[tauri::command]
 pub async fn session_rename(
     state: State<'_, AppState>,
+    app: tauri::AppHandle,
     session_id: String,
     title: Option<String>,
 ) -> Result<()> {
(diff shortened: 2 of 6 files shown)


yicheng47:
Review findings:

1. Medium: topbar Archive fails for running chats. `src/pages/RunnerChat.tsx:510` calls `api.session.archive()` directly, but `src-tauri/src/commands/session.rs:271` rejects rows where `status = 'running'`. The new kebab is rendered during running chats, so clicking Archive shows the archiving state and then errors instead of archiving. Mirror Sidebar’s flow: kill first when running, then archive.\n\n2. Medium: topbar Pin/Rename does not refresh the sidebar session row. `src/pages/RunnerChat.tsx:476` and `src/pages/RunnerChat.tsx:490` only call `refreshChatMeta()`. The backend `session_pin` / `session_rename` commands do not emit an event, and Sidebar only refreshes direct sessions on `session/exit`, `runner/activity`, or `session/archived` at `src/components/Sidebar.tsx:279`. Result: the header updates, but the sidebar pin/title stays stale until some unrelated refresh.\n\nVerification:\n- `pnpm exec tsc --noEmit` passed.\n- `pnpm run lint` passed with the existing Fast Refresh warning.\n- `cargo test --lib` passed.\n- `git diff --check origin/main...origin/fix/misc-followups` passed.\n\nGitHub Frontend is green; Backend was still in progress when I checked.

PR #64, opened 2026-05-06
fix: Shift+Enter newline + Cmd+V image paste in embedded xterm

### LINKED ISSUE #53: bug: Shift+Enter doesn't insert a newline in the embedded xterm
## Description

Inside the in-app terminal (xterm.js), pressing Shift+Enter does not insert a literal newline into the running TUI's input box. It behaves identically to a plain Enter — submitting whatever the agent CLI considers its current input.

This is the standard xterm.js default: both Enter and Shift+Enter fall through to `\r`. Most terminal emulators (iTerm2, VS Code, ghostty, etc.) ship a custom binding that sends `\x1b\r` (ESC+CR) on Shift+Enter, which TUIs like claude-code and codex interpret as \"insert newline in the input frame.\"

Without that binding, users cannot compose multi-line messages from the embedded terminal — they have to use the agent's CLI-specific fallback (e.g. claude-code's `\\` + Enter), which is awkward and easy to forget.

## Expected behavior

- **Enter** → submit (existing behavior, unchanged).
- **Shift+Enter** → insert a literal newline in the agent's input box.

The conventional way to express this on a PTY is to send `\x1b\r` (ESC followed by CR). Both claude-code and codex interpret that as a newline in their input frames.

## Steps to reproduce

1. Open a runner session (mission slot tab or direct chat).
2. Click into the xterm.
3. Type \"hello\".
4. Press Shift+Enter.
5. Observed: the line is submitted as if Enter was pressed. Expected: \"hello\" remains in the input frame with a newline appended, and the cursor moves to the next line of the input box.

## Relevant code

- `src/components/RunnerTerminal.tsx:180-186` — `term.onData` forwards every byte xterm.js emits straight to `session.injectStdin`. There is no `attachCustomKeyEventHandler` to intercept Shift+Enter.

## Fix sketch

Add a custom key event handler that maps Shift+Enter to `\x1b\r` and lets every other key fall through to the default xterm behavior. Roughly:

```ts
term.attachCustomKeyEventHandler((e) => {
  if (e.type !== \"keydown\") return true;
  if (e.key === \"Enter\" && e.shiftKey && !e.ctrlKey && !e.altKey && !e.metaKey) {
    const sid = sessionIdRef.current;
    if (sid && !disabledRef.current) {
      void api.session.injectStdin(sid, \"\\x1b\\r\").catch((err) => {
        onErrorRef.current?.(String(err));
      });
    }
    return false; // suppress the default \\r emission
  }
  return true;
});
```

Should be paired with `RunnerTerminal`'s existing `onData` so the default Enter path is untouched. Apply once in the same `useEffect` that constructs the terminal.

## Environment

- OS: macOS
- Version: 0.1.1

yicheng47:
## Summary

Two related xterm input-config fixes that bring the embedded terminal closer to what users expect from iTerm2/Ghostty/etc when running claude-code or codex.

### 1. Shift+Enter inserts a newline (closes #53)
- Add `attachCustomKeyEventHandler` in `RunnerTerminal` to map Shift+Enter to `\x1b\r` (ESC+CR), which claude-code and codex interpret as "insert newline in input frame".
- Plain Enter still falls through to the default `\r` via `term.onData`, so submission behavior is unchanged.

### 2. Cmd+V forwards clipboard images to the agent CLI
claude-code and codex already attach clipboard images natively when they see Ctrl+V (`\x16`) on stdin — they shell out to `pbpaste -Prefer png` (or the platform equivalent) and read the OS clipboard themselves. On macOS users instinctively press **Cmd+V**, which the WebView grabs as a *text* paste; xterm.js's default handler drops the image and pastes the placeholder text rep, so the keystroke that would have triggered the CLI's clipboard read never reaches the PTY.

- Add a capture-phase `paste` listener on `term.textarea`. When `clipboardData.items` contains an `image/*` file, `preventDefault` xterm.js's text-paste handler and inject `\x16` into the PTY. The CLI does the rest.
- Pure-text pastes fall through to xterm.js's default behavior unchanged.

## Test plan
- [ ] Open a runner session, type "hello", press Shift+Enter — line stays in the input frame with cursor on a new line.
- [ ] Press Enter normally — message submits as before.
- [ ] Confirm Ctrl+Enter / Alt+Enter / Cmd+Enter still pass through unchanged.
- [ ] Take a clipboard screenshot (Cmd+Shift+Ctrl+4), focus the xterm, press **Cmd+V** → claude-code shows `[Image #1]` placeholder in its input.
- [ ] Plain text Cmd+V → unchanged xterm text paste.
- [ ] Ctrl+V with an image still works (xterm forwards `\x16` natively).

### DIFF
--- a/src/components/RunnerTerminal.tsx
+++ b/src/components/RunnerTerminal.tsx
@@ -185,6 +185,29 @@ export function RunnerTerminal({
       });
     });
 
+    // Shift+Enter → ESC+CR so claude-code/codex insert a newline in their
+    // input frame instead of submitting. Plain Enter falls through to the
+    // default \r emission via onData above.
+    term.attachCustomKeyEventHandler((e) => {
+      if (e.type !== "keydown") return true;
+      if (
+        e.key === "Enter" &&
+        e.shiftKey &&
+        !e.ctrlKey &&
+        !e.altKey &&
+        !e.metaKey
+      ) {
+        const sid = sessionIdRef.current;
+        if (sid) {
+          void api.session.injectStdin(sid, "\x1b\r").catch((err) => {
+            onErrorRef.current?.(String(err));
+          });
+        }
+        return false;
+      }
+      return true;
+    });
+
     const pushSize = () => {
       const t = termRef.current;
       const sid = sessionIdRef.current;
--- a/src/components/RunnerTerminal.tsx
+++ b/src/components/RunnerTerminal.tsx
@@ -198,7 +198,7 @@ export function RunnerTerminal({
         !e.metaKey
       ) {
         const sid = sessionIdRef.current;
-        if (sid) {
+        if (sid && !disabledRef.current) {
           void api.session.injectStdin(sid, "\x1b\r").catch((err) => {
             onErrorRef.current?.(String(err));
           });
@@ -224,7 +224,7 @@ export function RunnerTerminal({
     // default behavior unchanged.
     const onPaste = (e: ClipboardEvent) => {
       const sid = sessionIdRef.current;
-      if (!sid) return;
+      if (!sid || disabledRef.current) return;
       const items = e.clipboardData?.items;
       if (!items) return;
       let hasImage = false;
--- a/src/components/RunnerTerminal.tsx
+++ b/src/components/RunnerTerminal.tsx
@@ -208,6 +208,43 @@ export function RunnerTerminal({
       return true;
     });
 
+    // Image paste support. claude-code and codex already know how to
+    // attach a clipboard image when they see Ctrl+V (\x16) on stdin —
+    // they shell out to `pbpaste -Prefer png` (or the platform
+    // equivalent) and read the OS clipboard themselves. The bug surface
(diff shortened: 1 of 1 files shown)


yicheng47:
Review note

`src/components/RunnerTerminal.tsx`: the new direct stdin paths need to respect the `disabled` terminal state after merging onto current `main`. `main` added `disabledRef` so stopped/resuming panes stop forwarding input and resize events; `onData` already checks it. These new handlers call `api.session.injectStdin` directly, so Shift+Enter or image Cmd+V can still write into a dead/resuming PTY and surface the session-not-found class of errors the prop was added to avoid.

Please guard both new paths when rebasing/merging:

```ts
if (!sid || disabledRef.current) return false; // Shift+Enter path
if (!sid || disabledRef.current) return;       // paste path
```

Checks run locally: `pnpm run lint` passed; `pnpm run build` passed, with the existing Node version warning in this shell (`18.20.8` vs project `>=22.12.0`). GitHub checks are passing for Backend and Frontend.

PR #66, opened 2026-05-06
docs: tmux session runtime plan + impl-doc renumbering

yicheng47:
## Summary
- New impl plan `docs/impls/0004-tmux-session-runtime.md` for moving session ownership from `portable-pty` to `tmux`. The plan is locked in against concrete prior art: code-level findings from `claude-squad/session/tmux/tmux.go` plus tmux man page / GitHub issues.
- Renumbers existing impl docs (`v0-mvp.md → 0001-…`, `crew-slots.md → 0002-…`, `direct-chats.md → 0003-…`) and updates inbound references across README, journals, and a handful of source-file doc pointers.

## What the plan locks in (vs. things easy to get wrong)
- **Private tmux server** (`-L runner`) — never share state with the user's `tmux ls` server; avoids the env-inheritance class of bugs.
- **Server-level options at first use** — `history-limit 50000`, `window-size manual`, `default-size 120x32`, `remain-on-exit on`. The last is the gate for `pane_dead_status` exit-code visibility (tmux#2552).
- **Exact-match `-t=name`** — `-t name` does prefix matching (claude-squad calls this footgun out inline at `tmux.go:459`). A target helper enforces the `=` form.
- **Capture flags pinned** — `-p -e` for live (xterm.js needs `-e`, must not have `-J`); `-S - -E -` only on attach replay. Polling cadence laddered 100/500/1000 ms with hash-debounce.
- **Input pipeline anti-pattern called out** — `claude-squad` keeps a hidden `tmux attach` PTY per session and writes user keystrokes to its stdin (`tmux.go:184,211-231`); that model is the proximate cause of cs#266 (dropped first prompts), cs#189/#216/#218 (capture races), and a 50ms stdin-nuking hack. We use `load-buffer -` → `paste-buffer -p -d` → `send-keys Enter` instead.
- **Quoting rules** — never `sh -c`; argv-only via `Command::new`; `--` separator before user payloads (tmux#1849); name regex `^[A-Za-z0-9_-]+$`.
- **Resize** — `resize-window -x cols -y rows` with a 50ms debounce, since no client is ever attached.
- **Status reconciliation** — `list-panes -F '#{pane_id} #{pane_dead} #{pane_dead_status} #{pane_pid} #{pane_current_command}'` is source of truth; `pane-died` hook is unreliable under simultaneous exits (tmux#2483) and used only as a wakeup.
- **Control mode (`tmux -CC`) explicitly deferred** — `%output` chunks split mid-UTF-8 codepoint (wezterm#6769); revisit when polling cost becomes measurable.
- **Windows fallback** — typed `RuntimeUnavailable::TmuxRequiresUnix`; native-pty runtime is the future Windows path. v1 ships macOS + Linux only.

## Test plan
- [ ] No code changes — docs-only PR. Verify the renumbered file links resolve in the GitHub-rendered view.
- [ ] CI passes (lint/fmt/tsc/clippy on doc-touching files only).
- [ ] Confirm the README and journal references point at the new `00NN-…` filenames.


yicheng47:
Review findings:

1. High: `capture-pane` snapshots are not a `session_output` byte stream. `docs/impls/0004-tmux-session-runtime.md` says to hash `capture-pane` output and emit `session_output` when it changes. That would append full screen snapshots into xterm.js, not replay incremental terminal bytes. Use `pipe-pane` or control mode for live raw output with `capture-pane` only for attach/replay, or add a separate snapshot event that resets/replaces the terminal view.

2. High: paste newline behavior is wrong. The plan uses `paste-buffer -p -d`, then says `\n` is not translated. The tmux 3.5a man page says `paste-buffer` replaces LF with CR by default. If we want literal multiline paste plus a separate submit Enter, use `paste-buffer -p -r -d ...`.

3. Medium: private `-L runner` does not prevent user tmux config from loading. Isolating the socket is good, but tmux still reads `~/.tmux.conf` when the private server starts. For deterministic app behavior, start the server with `-f /dev/null` or a generated Runner config, then apply our options.

4. Medium: `list-panes` target shape is off. Without `-s`, `list-panes` treats the target as a window. Use `list-panes -s -t=runner-<session_id>` or target `=runner-<session_id>:main`.

5. Medium: pane id capture can be simpler and safer. Use `new-session -d -P -F '#{pane_id}' ...` instead of a second `display-message` command. Also note that tmux `shell-command` is a shell command string, not argv, so the generated launch script path needs shell quoting if it can contain spaces.

Verification:
- `git diff --check` passed.

yicheng47:
Review findings:

1. High: server options may not apply to the first pane. `docs/impls/0004-tmux-session-runtime.md` says to run standalone `set-option` commands before any pane exists, but with an empty tmux server and default `exit-empty`, those settings can be lost when the command exits. This matters most for `history-limit`, which must be set before the first pane is created. Use a generated Runner tmux config via `-f <runner-tmux.conf>`, or run `set-option ... ; new-session ...` in one tmux command sequence so the options exist before the initial window/pane is created.

2. High: the FIFO `pipe-pane` attach order can stall the agent. The plan pipes to `cat >> output.fifo`, but opening a FIFO for write blocks until a reader is attached. The attach order starts `pipe-pane` before the forwarder task opens the FIFO, so output can back up and block the pane if replay/capture delays or fails. Open the FIFO reader first, then start `pipe-pane`, and shell-quote the FIFO path because `pipe-pane` also takes a shell command string. Also clarify that `-O` means output direction, not “open”; lowercase `-o` is the “only open if no pipe exists” flag.

3. Medium: the summary still contradicts the fixed stream design. The approach still says output is streamed by polling/capturing pane deltas, and the prior-art summary still describes v1 as polling `capture-pane`. Step 6 now correctly uses `pipe-pane` for live bytes and `capture-pane` only for replay. The summary should match that so implementation does not follow the stale path.

Verification:
- Reviewed PR #66 at `1610c7a`.

yicheng47:
Review findings:

1. High: FIFO is opened but not drained during replay. `docs/impls/0004-tmux-session-runtime.md` opens the FIFO reader, starts `pipe-pane`, then runs `capture-pane` and emits replay before starting the forwarder. If the agent writes enough output during capture/replay, the FIFO can fill and block `cat`, which then blocks the pane. Start the forwarder immediately after `pipe-pane`, buffer live chunks in memory until `session_replay` is emitted, then flush them as `session_output`. Also the note saying opening a FIFO read-only does not block without `O_NONBLOCK` is wrong; read-only open can block without `O_NONBLOCK`.

2. Medium: attach replay does not handle alternate-screen TUIs. The plan always uses `capture-pane -S - -E -`, which captures history/main screen. If Claude/Codex is in alternate screen, replay after reload can show stale or wrong content. Query `#{alternate_on}`; when true, capture current UI with `capture-pane -a -p -e` or define a two-part replay strategy.

3. Medium: some tmux examples still omit the generated config. Step 2 says every invocation must use `-L runner -f <runner.conf>`, but examples for `has-session`, `resize-window`, and `send-keys` omit `-f`. The helper example also omits `config_path`. Either update all snippets to use `tmux_cmd()` or mark them abbreviated.

Verification:
- Reviewed PR #66 at `693d640`.

yicheng47:
Review findings:

1. High: `pipe-pane -O -o` can break reattach after an app crash/restart. `docs/impls/0004-tmux-session-runtime.md` uses lowercase `-o`, meaning “only open if no pipe is currently open.” But the plan also expects reattach/app restart to recreate the live stream. If a stale pipe is still registered from the previous Runner process, `-o` will refuse to replace it, so live output may keep going to a dead FIFO or nowhere. On attach, explicitly close any existing pipe, then install the new pipe without `-o`; tmux already closes an existing pipe before running a new command.

2. Medium: generated config is not reloaded for an already-running Runner tmux server. The plan says the config is overwritten when stale, but `-f <runner.conf>` is only read when the tmux server starts. Since `exit-empty off` keeps the server alive, app restart/update can leave the server running with old options. Add an app-start reconciliation step: `source-file <runner.conf>` or explicit `set-option` before spawning new panes, ideally with a `@runner_config_version` check.

3. Medium: integration tests should use the generated config too. The tests mention only `-L runner-test-<pid>`. Without `-f <temp-runner.conf>`, tests can load user config and may not set `remain-on-exit`, so `pane_dead_status` assertions can be invalid.

Verification:
- Reviewed PR #66 at `5a0bacd`.

Issue #65, opened 2026-05-06
bug: direct chat fails with 'spawn claude: not found' when login-shell PATH resolution silently fails


yicheng47:
## Description

Clicking **Chat now** on a runner whose `command = "claude"` (or any bare CLI name) intermittently fails with:

```
session_start_direct: spawn claude: Unable to spawn claude because it doesn't exist on the filesystem and was not found in PATH
```

Root cause: the Runner GUI app is launched by launchd, which hands GUI apps a stripped PATH (`/usr/bin:/bin:/usr/sbin:/sbin`) that doesn't contain locations like `~/.local/bin`, `/opt/homebrew/bin`, etc. PR #36 (commit 982a9f2) addressed this by capturing the user's login-shell PATH at app startup via `$SHELL -ilc 'printenv PATH'` and merging it into every PTY's PATH (`SessionManager.shell_path`).

That capture is best-effort with **three silent-failure modes** and **no retry for the app lifetime**:

1. 2-second hard timeout (`RESOLVE_TIMEOUT` in `src-tauri/src/shell_path.rs:36`). A real `.zshrc` with oh-my-zsh / nvm / mise / asdf can spike past 2s on cold start.
2. Non-zero exit from the shell — silently returns `None`.
3. Markers missing from stdout — silently returns `None`.

Any of those returns `None`, then for the entire app lifetime PATH falls back to launchd's stripped default and `claude` (and `codex`, `mise`, etc.) become unspawnable. There is no log file, no UI surface, and no retry — restarting the app is the only recovery, and only works if the next cold-start resolution happens to come in under the timeout.

Made worse by PR #51 (commit ecf50e5): direct chats no longer get the bundled-bin dir on PATH, so `shell_path` is the **only** PATH source beyond launchd's stripped default for direct chats.

## Expected behavior

`Chat now` should reliably resolve `claude` regardless of:
- whether the first cold-start shell init exceeded 2 seconds,
- whether `$SHELL -ilc` exited non-zero on that boot,
- and without requiring the user to restart the app or know the absolute path.

When it can't, the error should explain *what was tried* (e.g., `claude: not found; checked PATH=…, fallback dirs ~/.local/bin, /opt/homebrew/bin, /usr/local/bin, ~/.cargo/bin`) so the failure isn't invisible.

## Steps to reproduce

Reliable repro is hard because the failure depends on a transient cold-start timeout. Synthetic repro:

[...]

PR #67, opened 2026-05-06
feat(session): SessionRuntime trait + tmux discovery foundation

yicheng47:
First slice of the tmux runtime migration spec'd in [`docs/impls/0004-tmux-session-runtime.md`](docs/impls/0004-tmux-session-runtime.md). Steps 1 + 2 of the plan: the trait the manager will delegate to, plus tmux binary discovery and the Runner-managed config writer.

## Summary

- **`src-tauri/src/session/runtime.rs`** (new) — `SessionRuntime` trait (`spawn` / `resume` / `stop` / `send_input` / `capture_since` / `resize`), supporting types (`SpawnSpec`, `RuntimeSession`, `CaptureCursor`, `CaptureChunk`), and typed `RuntimeError` (`TmuxRequiresUnix`, `TmuxNotFound { searched }`, `TmuxFailed { command, status, stderr }`, `Io`, `Msg`) with `From` into `crate::error::Error` so command code can `?` across the boundary.
- **`src-tauri/src/session/tmux.rs`** (new) — `resolve_tmux_binary()` searches `RUNNER_TMUX` → process PATH → `/opt/homebrew/bin` → `/usr/local/bin` → `/usr/bin`. Returns `TmuxRequiresUnix` on Windows so the runtime registry can refuse to construct `TmuxRuntime` cleanly. `write_runner_config()` idempotently writes `<app_data>/tmux.conf` with `exit-empty off`, `history-limit 50000`, `window-size manual`, `default-size 120x32`, `remain-on-exit on`, `status off`, `@runner_config_version "v1"`. `tmux_cmd()` pre-binds `-L <label> -f <config>` so individual call sites can't forget the global flags.
- **`src-tauri/src/session/mod.rs`** — exposes the two new modules.

Both new modules carry `#![allow(dead_code)]` because `manager.rs` doesn't reference them yet — that wiring lands in Step 5 of the plan, after the runtime metadata schema migration (Step 3) and the launch-script wrapper (Step 4). Wanted to keep this PR small and reviewable.

## Why a config file rather than runtime `set-option`

The plan's Step 5 spent some review rounds on this. With default `exit-empty=on` and an empty server, a stand-alone `tmux set-option -g …` invocation can hit a server that exits the moment the command returns, throwing the option away before the next `new-session`. `history-limit` is the worst offender — it must be set before the first pane is created. Loading from `-f <runner.conf>` makes the options apply at every server start, no race window. `exit-empty off` in the config keeps the server alive across Runner restarts so reattach works.

## Test plan

- [x] `cargo build -p runner` clean
- [x] `cargo test session::tmux` — 6 unit tests pass: `RUNNER_TMUX` honoring (executable check), bogus `RUNNER_TMUX` fall-through, idempotent rewrites, stale-content overwrite, required-options sanity, `tmux_cmd` argv shape
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --lib --tests -- -D warnings` clean
- [ ] No frontend changes; no UI / packaged-app smoke needed at this slice

## Follow-ups (separate PRs)

- Step 3: schema migration for `runtime_*` columns on `sessions`
- Step 4: launch-script wrapper + composed PATH (the actual fix for issue #65)
- Step 5: tmux `new-session` spawn + resize
- Step 6: `pipe-pane` + `capture-pane` streaming
- Step 7: input via `load-buffer` + `paste-buffer -p -r -d` + `send-keys`
- Step 8: resume + status + config reconciliation (`source-file` on stale `@runner_config_version`)
- Step 9: replace `portable-pty` in `manager.rs`
- Step 10: integration tests over `-L runner-test-<pid> -f <temp.conf>`


yicheng47:
Review findings:

1. Medium: the runtime trait bakes in the old output model before tmux is wired. `src-tauri/src/session/runtime.rs:166` says `capture_since` can be a `pipe-pane` drain or `capture-pane` snapshot and the manager should forward `CaptureChunk.bytes` as `session/output`. The tmux plan needs those separated: `pipe-pane` is live stream, while `capture-pane` is attach replay and must be emitted as a distinct replay payload. If this trait becomes the manager boundary as-is, the next chunk can reintroduce duplicated/stale terminal output. Split this now into explicit stream vs replay output, or return a discriminated event type.

2. Medium: `send_input` does not carry input intent. `src-tauri/src/session/runtime.rs:157` leaves the runtime to infer paste vs literal bytes vs control keys from "content / call site," but the call-site signal is not passed. Tmux needs different APIs for prompt paste, raw xterm bytes, and keys like Enter/C-c. This should be explicit now, either with `InputIntent` or separate methods, to avoid line-submit and prompt-buffer bugs later.

3. Low: the tmux resolver tests mutate process-global env vars in parallel tests. `src-tauri/src/session/tmux.rs:206` and `src-tauri/src/session/tmux.rs:224` set `RUNNER_TMUX` / `PATH`, then restore them, but Rust tests run concurrently by default. These can race with each other and become flaky. Use a test mutex/serial test, or refactor resolution behind an env provider.

Verification:
- `cargo test -p runner session::tmux` passed.
- `cargo clippy -p runner --lib --tests -- -D warnings` passed.
- `cargo test -p runner` passed: 169 tests.

yicheng47:
Review findings:

1. Medium: launch scripts render env var names without validation. `src-tauri/src/session/launch.rs:169` writes `export {k}=...` directly, but runner env is user/API-supplied via `src-tauri/src/commands/runner.rs:39`. An invalid key like `FOO-BAR` will make bash exit under `set -e` before the agent starts, and a crafted key can also break the script syntax. Validate env names as shell identifiers before persisting/rendering, and have `render_launch_script` fail cleanly for existing bad DB rows.

2. Low: the impl doc is internally inconsistent. The progress table says Step 1 is refined to `OutputStream` plus `Replay`/`Stream` and split input methods, but the Step 1 code block still shows the old `send_input` / `capture_since` API at `docs/impls/0004-tmux-session-runtime.md:141`. The next chunks should not follow that stale snippet.

Verification:
- `cargo test -p runner session::launch` passed.
- `cargo fmt --all --check` passed.
- `cargo clippy -p runner --lib --tests -- -D warnings` passed.
- `cargo test -p runner` passed: 181 tests.

yicheng47:
Review findings:

1. Medium: user env can still override the composed PATH in the generated launcher. `src-tauri/src/session/launch.rs:201` exports the deterministic `PATH`, then `src-tauri/src/session/launch.rs:205` exports every `script.env` entry afterward. Since `PATH` is a valid env name and `src-tauri/src/commands/runner.rs:165` only validates shell identifier shape, a runner env map containing `PATH` will replace the wrapper’s composed PATH. That weakens the main Step 4 guarantee and can bring back GUI-launch CLI resolution failures. Either reject reserved keys like `PATH` from runner env, filter it out when building `LaunchScript.env`, or render system PATH after user env so it always wins.

The prior findings look fixed:
- Invalid env names now fail at runner create/update and at launch render time.
- The Step 1 doc snippet now matches the actual `RuntimeOutput` / split-input trait shape.

Verification:
- `cargo test -p runner session::launch` passed: 15 tests.
- `cargo test -p runner commands::runner` passed: 18 tests.
- `cargo fmt --all --check` passed.
- `cargo clippy -p runner --lib --tests -- -D warnings` passed.
- `cargo test -p runner` passed: 185 tests.

yicheng47:
Review findings:

1. Medium: output can still be lost during spawn/resume attach. `src-tauri/src/session/tmux_runtime.rs:251` emits the replay snapshot before installing `pipe-pane`, and resume does the same at `src-tauri/src/session/tmux_runtime.rs:299`. Any bytes written after `capture-pane` completes but before `pipe-pane` attaches are neither in the replay nor in the live stream. The plan’s attach order is install/drain pipe first, capture replay, then flush buffered stream bytes.

2. Medium: the FIFO forwarder can leak blocked threads. `src-tauri/src/session/tmux_runtime.rs:465` opens the FIFO read/write, so this process keeps a writer fd alive forever. That means `src-tauri/src/session/tmux_runtime.rs:476` will not see EOF when tmux closes the pipe, and if the receiver is dropped while no bytes arrive, the thread remains blocked in `read()`. Repeated resume/detach paths can accumulate stuck threads.

3. Medium: failed post-spawn setup leaks a tmux session. After `new-session` succeeds at `src-tauri/src/session/tmux_runtime.rs:228`, failures from `emit_replay` or `install_pipe_pane` return an error without killing the newly created tmux session. That leaves a live agent pane with no persisted `RuntimeSession` for the manager to reconcile.

The previous PATH override finding is fixed in `5e1fc32`.

Verification:
- `cargo test -p runner session::tmux_runtime` passed.
- `cargo test -p runner session::launch` passed.
- `cargo test -p runner session::tmux_runtime -- --ignored` passed: 3 tmux integration tests.
- `cargo fmt --all --check` passed.
- `cargo clippy -p runner --lib --tests -- -D warnings` passed.
- `cargo test -p runner` passed: 193 passed, 3 ignored.

yicheng47:
Review findings:

1. High: current head fails `cargo test -p runner session::tmux_runtime`. `src-tauri/src/session/tmux_runtime.rs:234` allows a live pane row like `"%0 0"` to parse with `pid = None`, even though the parser docs and test require `pane_id`, `pane_dead`, and `pane_pid` at minimum. The failing test is `parse_pane_status_line_rejects_bad_shapes` at `src-tauri/src/session/tmux_runtime.rs:981`.

2. Medium: the new status reconciliation path still does not prove dead-pane detection works. `cargo test -p runner --lib session::tmux_runtime::tests::integration_status_alive_then_dead -- --ignored` fails with `pane never went dead`. That means `status()` at `src-tauri/src/session/tmux_runtime.rs:553` is not observing `pane_dead=1` / `pane_dead_status` for a quick-exit command, so Step 8 reconciliation would not reliably mark exited sessions stopped/crashed.

Verification:
- `cargo test -p runner session::launch` passed.
- `cargo test -p runner --lib session::tmux_runtime::tests::integration_spawn_and_observe -- --ignored` passed.
- The broader tmux runtime test target currently fails as noted above.

yicheng47:
Review finding:

1. Medium: `reconcile_config()` treats a missing `@runner_config_version` as a no-op. `src-tauri/src/session/tmux_runtime.rs:104` returns `Ok(false)` for any failed `show-options`, including the exact legacy-server case the docs say should reload. The plan says missing/different stamp should `source-file` the config, but this path skips reload, so old tmux servers can keep stale options.

Verification:
- The two previously posted findings still reproduce on head `13c4f76`: `cargo test -p runner session::tmux_runtime` fails, and the ignored `integration_status_alive_then_dead` test still cannot observe the pane dead state.

yicheng47:
Review findings:

1. High: app restart still throws away live tmux sessions. `src-tauri/src/lib.rs:66` bulk-marks every `running` session row as `stopped` on startup, and there is no production caller of `SessionRuntime::resume`. When the user later resumes that row, `SessionManager::resume` at `src-tauri/src/session/manager.rs:946` calls `runtime.spawn`, and `TmuxRuntime::spawn` at `src-tauri/src/session/tmux_runtime.rs:338` kills the existing `runner-<session_id>` tmux session before starting a new one. So closing/reopening Runner does not attach to the preserved tmux pane; it marks it stopped, then resume destroys it.

2. Medium: post-spawn resize failure leaks an untracked tmux session. `src-tauri/src/session/tmux_runtime.rs:392` runs `resize-window` after `new-session`, but before the cleanup-wrapped `attach_streaming` block. If resize fails, `spawn()` returns `Err` without `kill-session`; the manager then rolls back/deletes the DB row, leaving the agent running in tmux with no tracked `RuntimeSession`.

Verification:
- `cargo test -p runner --lib session::tmux_runtime -- --ignored` passed.
- `cargo clippy -p runner --lib --tests -- -D warnings` passed.

The previous parser/status/reconcile findings look fixed on `abc133b`.

yicheng47:
Review findings:

1. Medium: app-start reattach restores PTYs but not mission routing. src-tauri/src/lib.rs:127 reattaches running sessions, while buses/routers are still empty at src-tauri/src/lib.rs:135. For mission sessions, an agent can keep running in tmux and append ask_lead, human_said, or runner_status before the workspace mounts mission_attach; later reconstruction explicitly does not replay stdin side effects at src-tauri/src/router/mod.rs:199. That means mission coordination can silently miss events after app restart. Either mount routers/buses for running missions during startup, or stop/mark mission sessions as needing explicit resume instead of letting them run detached.

2. Medium: failed reattach can strand a live tmux pane. src-tauri/src/session/manager.rs:1424 marks the DB row stopped if attach_existing fails, but the pane was already reported alive and is never stopped. Since attach_existing can fail at src-tauri/src/session/manager.rs:1469 while installing the output stream, the agent may keep running in tmux while the UI/DB says stopped.

Verification:
- git diff --check origin/main...HEAD passed.
- cargo test -p runner passed.
- cargo test -p runner --lib session::tmux_runtime -- --ignored passed.
- cargo clippy -p runner --lib --tests -- -D warnings passed.

The previous two findings about startup bulk-stop and resize cleanup look fixed on the latest head.

yicheng47:
Review findings:

1. Medium: fresh spawns can now lose startup output. src-tauri/src/session/tmux_runtime.rs:780 skips the capture-pane snapshot for fresh sessions, but the process starts at new-session before pipe-pane is installed at src-tauri/src/session/tmux_runtime.rs:771. Any banner, prompt, or fast command output emitted in that window is no longer recovered. The existing test helper even documents this race at src-tauri/src/session/tmux_runtime.rs:1128, then avoids it by repeating output, so the current tests won’t catch this regression.

2. Medium: kill can still mark/archive a session even if tmux did not actually stop it. src-tauri/src/session/tmux_runtime.rs:483 discards kill-session failure, while src-tauri/src/session/manager.rs:1365 forces the forwarder to exit and reconcile the row. If tmux fails to kill the session, the DB can say stopped/archived while the agent keeps running.

3. Medium: mission crash state is still lost on app restart. src-tauri/src/session/manager.rs:1478 handles mission rows before checking runtime.status, so a dead mission pane with non-zero exit is marked stopped instead of reaching the crashed reconciliation path.

Verification:
- git diff --check origin/main...HEAD passed.
- cargo test -p runner passed.
- cargo test -p runner --lib session::tmux_runtime -- --ignored passed.
- cargo clippy -p runner --lib --tests -- -D warnings passed.

yicheng47:
Review findings:

1. Medium: failed kill still reconciles the row as stopped. src-tauri/src/session/manager.rs:1364 captures stop_result, but src-tauri/src/session/manager.rs:1370 still flips the forwarder stop flag and joins before returning the error. The forwarder then breaks, sees was_killed, and writes stopped at src-tauri/src/session/manager.rs:1175. So if tmux refuses to stop, the API returns an error, but the live handle is gone and the DB can still say stopped. The same “ignore stop failure then mark stopped” shape remains for live mission startup cleanup at src-tauri/src/session/manager.rs:1492.

2. Medium: fresh-spawn output loss is reduced, not eliminated. src-tauri/src/session/tmux_runtime.rs:366 starts the process, then src-tauri/src/session/tmux_runtime.rs:381 attaches pipe-pane as the next tmux command. That shrinks the race, but any command that prints/exits immediately before pipe-pane attaches is still lost because fresh spawns still skip the snapshot at src-tauri/src/session/tmux_runtime.rs:830. This matters for fast shell/custom runners, even if Claude/Codex startup is usually slower.

Verification:
- git diff --check origin/main...HEAD passed.
- cargo test -p runner passed.
- cargo test -p runner --lib session::tmux_runtime -- --ignored passed.
- cargo clippy -p runner --lib --tests -- -D warnings passed.

The mission crash-state reconciliation finding looks fixed.

Issue #50, opened 2026-05-04
harden first-turn stdin injection: replace fixed 2.5s timer with PTY read-back verification


yicheng47:
## Background

#45 fixed the codex system-prompt disruption by (a) defaulting bypass-permission flags so the startup dialog is out of the boot sequence, and (b) routing the brief through stdin injection — parallel to claude-code's existing path. Both runtimes now share the same delivery mechanism in \`schedule_first_prompt\` (\`src-tauri/src/session/manager.rs:1582\`).

That mechanism is a hand-rolled fixed timer:

\`\`\`rust
// src-tauri/src/session/manager.rs:1552
const FIRST_PROMPT_DELAY: Duration = Duration::from_millis(2500);
\`\`\`

On a fresh spawn we sleep 2.5s, type the body, sleep 80ms, type Enter. The same pattern exists at \`router/handlers.rs:29\` as \`LEAD_LAUNCH_PROMPT_DELAY\` for the lead-launch dispatch race.

## Problem

The 2.5s budget is tolerable today because the bypass flags make startup deterministic — TUI render + raw-mode stdin bind, no human-in-the-loop dismissal. But:

1. Users who toggle bypass OFF (sensitive/sandboxed setups) regress to the original race: the dialog is still up when the timer fires, the brief lands behind it.
2. Future codex / claude-code versions with slower first render (async sandbox init, network-bound token refresh, plugin auto-load) could push past 2.5s. Failure is silent: bytes typed before the editor binds get echoed once and lost; the agent boots with no context and the user just sees an empty input.
3. There is no \"the TUI is ready\" signal — we trust wall-clock time.

## Proposed direction (option 2 from the #45 design discussion)

Replace the fixed timer with a read-back verification loop:

1. After spawn, watch the PTY's output stream for evidence that the editor is bound (echo of a probe byte typed into stdin, or a known input-box render marker).
2. Type the brief, then continue watching for the brief's own bytes to echo back inside the input region within a budget (e.g. 5–8s).
3. If the read-back lands, type Enter. If it doesn't, retry once, then surface a structured error to the workspace feed instead of leaving the agent silently context-less.

Generalize the helper so both \`schedule_first_prompt\` and the \`LEAD_LAUNCH_PROMPT_DELAY\` site share it — they're the same race in two places.

## Acceptance

- \`schedule_first_prompt\` no longer references \`FIRST_PROMPT_DELAY\` as the only gate; instead it waits on observed PTY readiness with a configurable upper bound.
[...]

PR #72, opened 2026-05-09
docs: plan 0005 — first-prompt readback verification

yicheng47:
## Summary
- Plan-only PR. Implementation lands in a follow-up commit on the same branch.
- Replaces the `FIRST_PROMPT_DELAY = 2500ms` blind-wait in `inject_first_turn` with a paste → `capture-pane` → verify → retry → Enter loop. Targets v0.1.4 release.
- Fixes the post-v0.1.3 race where the persona occasionally fails to land in the agent's input box on fresh chat spawn (issue #50, also reported live).

## What's in the plan

- **Why**: 2500ms blind wait isn't enough under contention — paste lands during Claude Code's pre-input handler (trust dialog / Node init / banner phase) and gets eaten. `claude-squad` (cs#266) ships with the same bug; we already do better, goal is to do better than "better".
- **Approach**: post-paste readback via tmux `capture-pane -p -e`. Match either the body marker (first ≤32 chars of trimmed body, line-bounded, codex + short claude-code paths) OR the literal `Pasted text` placeholder (Claude Code's long-paste placeholder). Either match → send Enter. Retry on miss.
- **Timing budget**: initial 1500ms / render 600ms / between 800ms / max 4 attempts. Best case 2100ms (faster than today). Worst case 7100ms (only worse if today would have failed anyway, in which case persona actually lands).
- **Touch surface**: 1 new trait method (`SessionRuntime::capture_visible`), 1 new manager method (`inject_paste_with_verify`), 2 helpers, 3 new unit tests. `cfg(test)` zeros all durations so unit tests stay synchronous.

## Test plan

- [ ] Implementation commit lands on this branch
- [ ] `cargo fmt` / `cargo clippy --lib --tests -- -D warnings` clean
- [ ] `cargo test` — 197 existing + 3 new pass
- [ ] Manual smoke: spawn 3 direct chats with personas in parallel, confirm all 3 show persona as first user turn

### DIFF
--- a/docs/impls/0005-first-prompt-readback.md
+++ b/docs/impls/0005-first-prompt-readback.md
@@ -92,33 +92,59 @@ exposed. The plan covers all three.
 A new manager-side method `inject_paste_with_verify` runs the loop:
 
 ```text
-sleep(initial_wait)          // first-cut readiness wait, smaller than today
+sleep(initial_wait)              // first-cut readiness wait, smaller than today
+before = runtime.capture_visible(session)
+marker = paste_marker(body)
+before_marker_count = count_substr(strip_ansi(before), marker)
+before_placeholder_count = count_placeholders(before)
 loop attempt = 0..max_attempts:
     if session no longer in self.sessions: bail (user killed it)
     runtime.paste(session, body)
-    sleep(render_wait)        // give tmux + agent TUI time to render
-    snapshot = runtime.capture_visible(session)
-    if pane_acknowledged_paste(snapshot, marker(body)):
+    sleep(render_wait)           // give tmux + agent TUI time to render
+    after = runtime.capture_visible(session)
+    after_stripped = strip_ansi(after)
+    if count_substr(after_stripped, marker) > before_marker_count:
+        runtime.send_key(session, "Enter")
+        return Ok
+    if body.len() >= PLACEHOLDER_MIN_BODY_LEN
+       && count_placeholders_stripped(after_stripped) > before_placeholder_count:
         runtime.send_key(session, "Enter")
         return Ok
     sleep(between_attempts)
-return Err("persona not visible after N attempts")
+return Err("paste not visible after N attempts")
 ```
 
-`pane_acknowledged_paste` matches **either** of:
+Both acceptance signals use a **count delta** between `before` (one
+capture, taken once before the loop) and `after` (one capture per
+attempt). Either delta increasing by ≥ 1 means our paste landed:
 
-1. The body marker (first ≤32 chars of trimmed body, line-bounded)
-   appears verbatim in the pane after CSI escapes are stripped. Covers
-   short personas (Claude Code shows the paste verbatim under its
-   wrap threshold) and the codex TUI (always verbatim).
-2. The literal substring `Pasted text` appears in the pane. Covers
-   long personas where Claude Code substitutes a `[Pasted text #N
-   +M lines]` placeholder for the actual content.
+1. **Body-marker count delta.** The first ≤32 chars of trimmed body
+   (line-bounded; stop at first `\n`) — `paste_marker(body)` —
+   counted as a substring in the CSI-stripped snapshot. Covers
+   short pastes (Claude Code shows them verbatim) and the codex
+   TUI (always verbatim regardless of length). For 8-byte bodies
+   like `"continue"` this is the only signal that fires; using the
+   delta (rather than just "marker is now visible") handles the
+   resume case where the prior conversation already contains the
+   marker text — a `continue` resume against a transcript
+   mentioning the word "continue" still verifies correctly,
+   because the paste pushes the count from N to N+1.
+2. **Placeholder count delta.** `count_placeholders` is occurrences
+   of the literal `Pasted text` substring in the CSI-stripped
+   snapshot. Only consulted when `body.len() >=
+   PLACEHOLDER_MIN_BODY_LEN` (64 bytes — well below Claude Code's
+   actual wrap threshold of ~200, but above any reasonable
+   short-prompt zone where the placeholder shouldn't fire). Skipping
+   for short bodies plus using a delta both defend against the
+   resume false-ack case the reviewer flagged: a resumed pane that
+   already shows `[Pasted text #5 ...]` from prior turns has
+   `before_placeholder_count = 1`; a failed `continue` paste leaves
+   `after_placeholder_count = 1`; delta = 0 → reject and retry.
 
-Either match means the paste landed in the input box. The placeholder
-match is Claude-Code-specific by design — codex is the other
-production runtime and it always shows verbatim, so the body-marker
-match covers it.
+The capture-before is taken **once**, before the retry loop. Each
+retry compares against that fixed baseline, so duplicate pastes that
+actually do land (a paste-then-render race resulting in two
+placeholders) still verify on the first delta-positive attempt.
 
 ### Timing budget
 
@@ -228,18 +254,24 @@ impl StdinInjector for SessionManager {
 }
 ```
 
-Update `Router::inject_and_submit_delayed` to drop the
-keystroke-then-`\r` chord on the non-zero-delay path and use
-`injector.inject_paste_with_verify(session_id, &body)` instead. The
-verified path delivers the body as a bracketed paste (so multi-line
-launch prompts render correctly inside the input box, not as
-character-by-character keystrokes that wrap differently) and
-internally handles the Enter once it's confirmed the body landed.
+Update `Router::inject_and_submit_delayed` so **both** branches route
+through `injector.inject_paste_with_verify(session_id, &body)`,
+regardless of whether the `delay` parameter is zero or non-zero. The
+delay parameter now controls only "spawn a thread vs run inline";
+the *routing* decision is settled by the method name (the
+`_delayed` suffix means launch-prompt-class injection, which always
+needs the verified path):
 
-The zero-delay path (used by router unit tests with
-`LEAD_LAUNCH_PROMPT_DELAY = ZERO`) keeps the existing inline-`inject`
-behavior — those tests assert byte-write counts against a fake
-injector and shouldn't see a different call shape.
+- **Non-zero delay** (production): spawn a thread, immediately call
+  `inject_paste_with_verify` (no outer `sleep(delay)` — see Risks /
+  budget below).
+- **Zero delay** (tests): call `inject_paste_with_verify` inline.
+  Under `cfg(test)` the verified path's own durations are all zero,
+  so this stays a synchronous millisecond no-op.
+
+Drop the trailing `\r` chord entirely from the delayed path — the
+verified primitive sends Enter internally once it confirms the body
+landed.
 
 `inject_and_submit` (the synchronous, non-delayed path used by
 `human_said` and `ask_lead`) stays on raw byte injection. Those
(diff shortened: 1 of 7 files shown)


yicheng47:
Review finding:

1. Medium: the plan should include Codex and the mission lead launch prompt path, not only `inject_first_turn`. Codex uses the same stdin first-turn path because startup permission/approval UI can swallow or misorder prompt argv (`src-tauri/src/router/runtime.rs:13-16`). The proposed change covers Codex direct chat and non-lead mission prompts through `inject_first_turn`, but lead mission startup still uses `Router::inject_and_submit_delayed` (`src-tauri/src/router/handlers.rs:81`), whose `StdinInjector` only supports raw byte injection (`src-tauri/src/router/mod.rs:39-46`). A Codex lead can hit the same startup-readiness race, and Claude lead can too. The plan should either extend the router injector with a verified paste+submit method or route lead launch prompt injection through `SessionManager::inject_paste_with_verify`.

yicheng47:
Good catch — the plan was missing the third injection site. Updated in `8314616`.

Confirmed by reading the code:

- `Router::inject_and_submit_delayed` (`src-tauri/src/router/mod.rs:425`) is the lead launch prompt path. Two callers: `handlers::mission_goal` (`handlers.rs:81`, mission boot) and `Router::fire_lead_launch_prompt` (`mod.rs:518`, resume-fresh-fallback when claude-code's conversation file is missing).
- It calls `StdinInjector::inject(session_id, body)` → `SessionManager::inject_stdin` → `tmux send-keys -l` (per-character keystrokes, *not* bracketed paste). So the lead path is actually *more* fragile than the direct-chat path it was meant to mirror — multi-line bodies become character-by-character keystrokes, which can wrap differently inside the agent's input box than a real paste, on top of the same readiness race.

Plan changes:

1. New "Three injection paths share the race" subsection under **Why** with a table mapping each path to its caller, today's primitive, and what it's used for.
2. New `src-tauri/src/router/mod.rs` section under **Touch surface**: extend `StdinInjector` with `inject_paste_with_verify`, swap `inject_and_submit_delayed`'s non-zero-delay path to use it (drops the `\r` chord — verified path handles Enter internally). Zero-delay path stays on raw `inject` so the `LEAD_LAUNCH_PROMPT_DELAY = ZERO` router unit tests don't break. `inject_and_submit` (synchronous, used by `human_said` / `ask_lead`) stays on raw byte injection because those target already-running agents.
3. New router-level unit test `lead_launch_prompt_routes_through_verified_paste` to guard the wiring.
4. Test stub note about updating the router's fake `StdinInjector` to support the new method.

Now covers Codex mission leads (most exposed — argv `[PROMPT]` gets swallowed by the approval dialog so they always go through stdin) and Claude Code mission leads, as well as the original direct-chat persona + worker preamble + resume-`continue` paths.

yicheng47:
Review findings:

1. Medium: `Pasted text` is a global success marker, which can false-ack `continue` on resume. The plan says `pane_acknowledged_paste` accepts either the body marker or `Pasted text` (`docs/impls/0005-first-prompt-readback.md:108`), then applies the same verified path to `schedule_continue_on_resume` (`docs/impls/0005-first-prompt-readback.md:195`). A resumed pane may already show an old Claude `[Pasted text ...]` placeholder, so a failed `continue` paste could still be treated as acknowledged. The placeholder match should be conditional, likely only for long bodies, or compare before/after capture.

2. Medium: the lead launch path may accidentally stack delays. The plan keeps `inject_and_submit_delayed` as the lead route and adds `inject_paste_with_verify` using `FIRST_PROMPT_CONFIG` (`docs/impls/0005-first-prompt-readback.md:200`), whose own initial wait is 1500ms (`docs/impls/0005-first-prompt-readback.md:123`). If the implementation keeps the old 2500ms outer sleep, lead prompts become 2500 + 1500 + render wait. The plan should state which layer owns readiness waiting.

3. Low: the proposed router test is underspecified. It says configure `LEAD_LAUNCH_PROMPT_DELAY` non-zero and fire `mission_goal` (`docs/impls/0005-first-prompt-readback.md:280`), but that constant is currently `ZERO` under `cfg(test)`. Add a test hook/config, or call the delayed helper directly with a non-zero duration and synchronize on the fake injector.

Codex/lead coverage from the previous comment is now addressed conceptually. `git diff --check origin/main...origin/fix/first-prompt-readback` passed.

yicheng47:
All three findings addressed in `ed1ca75`.

**1. `Pasted text` false-ack on resume → count-delta scheme.** Plan now captures pane state *once* before the retry loop and stores `before_marker_count` + `before_placeholder_count` as baselines. Each post-paste capture compares: accept on `marker_count > before_marker_count` OR (when `body.len() >= PLACEHOLDER_MIN_BODY_LEN = 64`) `placeholder_count > before_placeholder_count`. A resumed pane that already shows `[Pasted text #5 ...]` from a prior turn has `before_placeholder_count = 1`; a failed `continue` paste leaves the after-count at 1; delta = 0, reject and retry. The marker check uses delta too (not "marker is now visible") so a `continue` resume against a transcript mentioning the word "continue" still verifies on a delta of 0 → 1.

**2. Stacked delay → verified path owns readiness.** New "Readiness budget — verified path owns it" subsection under the router touch-surface. `inject_and_submit_delayed`'s thread immediately calls `inject_paste_with_verify` with no outer sleep; the verified path's own `initial_wait` (1500ms) + `render_wait` (600ms) is the only pre-paste budget. Net lead-path production cost: ~2100ms best case (faster than today's blind 2500ms), 7100ms worst case (vs today's 2500ms-then-fail). The legacy `LEAD_LAUNCH_PROMPT_DELAY` constant becomes a vestigial argument; plan accepts either keeping it for one release as a no-op or deleting it in the same commit (implementer's call).

**3. Router test underspecified → routing not keyed on delay.** Restructured routing: `inject_and_submit_delayed` always routes through `inject_paste_with_verify` regardless of delay value. The `delay` parameter now controls only "spawn thread vs run inline"; the routing decision is settled by the method name. Test runs cleanly under the existing `cfg(test)` zero-delay shape — no test hook or non-zero override needed.

Bonus addition: **Existing test impact** section enumerating which router tests' assertions migrate. `RecordingInjector` gains an `InjectKind` enum (`Raw` / `PasteVerified`) tagged on each recorded entry; existing helpers project the kind away, so existing body-content assertions (`tests.rs:230`, `:273`, `:296`) stay valid as-is. New `paste_pushes_for(...)` helper for the new lead-routing test.

Self-found follow-on while addressing #1: my first draft of the marker check used "marker visible AND not in before" (presence gate), which would have broken `continue`-on-resume when the prior conversation contained the word "continue" anywhere. Switched marker check to also use count-delta — same scheme as placeholder, just on a different needle.

Risks section updated to cover the *dynamic* false-positive case (agent's banner happens to render persona-shaped text between before and after captures): vanishingly rare for "You are a/an…" personas, impossible for codex (fixed banner). Mitigation pointer if it ever bites: scope marker counting to bottom-N rows where the input box lives.

yicheng47:
Review findings:

1. Medium: the new pre-capture step has undefined error handling. The loop captures `before` before any retry at `docs/impls/0005-first-prompt-readback.md:95`, but later says `InertRuntime` returns `Err` and the verify loop treats capture errors as “not seen” and retries at `docs/impls/0005-first-prompt-readback.md:310`. The plan should state whether a failed baseline capture becomes zero counts or aborts.

2. Medium: Codex long prompts may still fail readback. The body marker is only the first <=32 chars at `docs/impls/0005-first-prompt-readback.md:121`, but `capture_visible` only sees the visible pane. For long Codex mission prompts, the editor may show the bottom/end of the pasted text, not the first line. Consider checking both first and last non-empty markers, or explicitly validating Codex’s long-paste display behavior.

3. Low: the test section is stale/inconsistent after the router test was added. It now lists three manager tests plus one router test, but rollout still says “three new tests + 197 existing” at `docs/impls/0005-first-prompt-readback.md:426`. Also the test cases still say `pane_content` while the fake was renamed to `pane_pre_paste` / `pane_post_paste`.

Previous Codex/lead, false-ack, stacked-delay, and router-test-delay issues look addressed conceptually. `git diff --check origin/main...origin/fix/first-prompt-readback` passed.

yicheng47:
All three round-3 findings addressed in `88495b9`.

**1. Pre-capture error handling → explicit "Capture error handling" subsection** under Approach:
- **Baseline (`before`) capture fails**: treat as zero counts and proceed. Implemented as `unwrap_or_default()` → empty `Vec<u8>` → `count_substr(b"", _) == 0`. The alternative (abort and fall back to raw send-keys) regresses every transient tmux flake to the pre-fix state, which is worse than a rare false-ack.
- **Per-attempt (`after`) capture fails**: skip *this* attempt's check, continue the loop after `between_attempts`. No inline retry — the loop's natural cadence rate-limits.
- Both cases log to stderr so operators can correlate.

**2. Codex long prompts → tail marker added.** Plan now defines `paste_markers(body) -> (head_marker, tail_marker)` and counts deltas on **both**:
- `head_marker` = first ≤32 chars of body's first non-empty line (covers verbatim-from-top renders — claude-code short pastes, codex short pastes).
- `tail_marker` = last ≤32 chars of body's last non-empty line (covers TUIs that scroll to keep the cursor visible — codex's chat composer is the known case where a multi-KB prompt shows its trailing lines instead of its leading lines).
- `Pasted text` placeholder count delta — claude-code-specific long-paste signal, gated on `body.len() >= 64`.

Three signals OR'd; any one delta ≥ 1 accepts. Added a manual-smoke item: "start a fresh codex-lead mission with a long preamble+brief (≥2KB), confirm the lead receives the launch prompt and submits it".

For short bodies where head and tail overlap, both signals fire on the same paste — same effect as "either marker matched", no double-counting risk.

**3. Test section stale → unified list of 5, naming consistent.**
- Manager-level: 4 tests (added `continue_resume_rejects_stale_placeholder` as a direct regression guard for the round-2 false-ack fix — `body = b"continue"`, `pane_pre_paste = pane_post_paste = "[Pasted text #5 +20 lines]"`, expected: `max_attempts` pastes, zero Enters).
- Router-level: 1 test (`lead_launch_prompt_routes_through_verified_paste`).
- All references to the old single-field `pane_content` updated to `pane_pre_paste` / `pane_post_paste`.
- Rollout updated: "5 new tests + 197 existing pass". Codex long-prompt manual smoke added.

Bonus: tightened Non-goals `InertRuntime` entry to point at the new error-handling subsection instead of restating it inline.

yicheng47:
Direction note:

I think this is the right direction. For this class of bug, `tmux paste-buffer` + `capture-pane` readback + retry is the right level of mechanism. Other tmux-based orchestrators mostly do something weaker: send prompt text, sleep briefly, then send Enter, sometimes with `capture-pane` scraping for status/trust prompts. That helps operationally, but it does not prove the first prompt landed before Enter.

Runner’s proposed path is stronger because it verifies that the pasted content appeared in the pane before submitting. The caveat is that `capture-pane` proves “visible in terminal,” not “accepted into model context,” but for Claude Code/Codex interactive TUIs that is the best practical proxy we have without a stable readiness/accepted-prompt API.

My preference: make verified paste-submit the single first-turn primitive for direct personas, worker preambles, lead launch prompts, and resume `continue`. Leave normal human/live-message injection on the simpler ready-session path.

yicheng47:
Aligned — that's exactly what the plan does:

| Path | Primitive |
|---|---|
| Direct-chat persona (`schedule_direct_first_prompt` → `inject_first_turn`) | `inject_paste_with_verify` |
| Non-lead worker preamble (`schedule_mission_first_prompt` → `inject_first_turn`) | `inject_paste_with_verify` |
| Lead launch prompt (`handlers::mission_goal` / `Router::fire_lead_launch_prompt` → `inject_and_submit_delayed`) | `inject_paste_with_verify` (via `StdinInjector` trait extension) |
| Resume `continue` (`schedule_continue_on_resume`) | `inject_paste_with_verify` |
| Live-message injection (`human_said` / `ask_lead` → `inject_and_submit`, no delay) | Raw `inject` (target is already-running, TUI bound — no readback needed) |

The "verified path is the sole owner of pre-paste readiness waiting" rule from round-2 also makes this clean: `inject_paste_with_verify` is the only thing in the codebase that knows about `FIRST_PROMPT_CONFIG` timing. The 80ms post-`inject` `\r` chord and the legacy `LEAD_LAUNCH_PROMPT_DELAY` outer sleep both go away on the four first-turn paths.

On the `capture-pane` caveat — agreed, "visible in terminal" is the best practical proxy without a real readiness API. The three-signal OR (head-marker / tail-marker / placeholder count delta) is a slight defense-in-depth on top of that proxy: if the input editor decides to scroll, swap to a placeholder, or do anything other than show the literal first 32 chars, we still detect the landing. Won't help if the agent silently ignores the paste while displaying it (no production cases known), but at that point we'd want runtime telemetry, not capture-pane.

Proceeding to implementation. Will keep the PR scope tight to these four paths.

PR #82, opened 2026-05-11
fix(terminal): image paste — restore NSPasteboard so agent sees real bytes

### LINKED ISSUE #79: bug: image paste into runner terminal not captured by agent CLI
## Description

Pasting an image into a runner session's terminal (xterm.js mounted by
\`RunnerTerminal\`) doesn't reach the underlying agent CLI as an
attached image. The intended flow is: Cmd+V → JS \`paste\` event with
\`image/*\` clipboard item → \`RunnerTerminal\` swallows xterm.js's
default text-paste and injects \`\\x16\` (Ctrl-V) on stdin → the agent
CLI (claude-code / codex) sees Ctrl-V and shells out to \`pbpaste
-Prefer png\` to read the OS clipboard. Reported behavior is that this
does not actually attach an image on the agent side.

## Expected behavior

Cmd+V (or Ctrl+V) in a focused runner terminal with an image on the
clipboard results in the agent CLI seeing the image as an attached
input — same as if the user had pressed Ctrl+V while running the CLI
directly in a host terminal.

## Steps to reproduce

1. Take a screenshot (Shift+Cmd+Ctrl+4 on macOS — copies to clipboard).
2. Open a runner chat with a live session running claude-code or codex.
3. Click into the embedded terminal so it has focus, then press Cmd+V.

Observed: agent does not pick up the image (or picks up garbled
placeholder text instead of the image). Expected: agent treats the
paste as an image attachment, same as when the CLI is run standalone.

## Relevant code

- \`src/components/RunnerTerminal.tsx:211-244\` — \`onPaste\` listener
  attached to xterm's hidden \`textarea\`. Detects images by iterating
  \`e.clipboardData.items\` and matching \`it.kind === "file" &&
  it.type.startsWith("image/")\`. On a hit, it
  \`preventDefault\`+\`stopImmediatePropagation\` and calls
  \`api.session.injectStdin(sid, "\\x16")\`.
- \`src/components/RunnerTerminal.tsx:246\` — listener registered with
  \`{ capture: true }\` on \`term.textarea\` to outrun xterm.js's own
  paste handler.
- \`src-tauri/src/session/manager.rs\` — backend \`inject_stdin\` path
  that delivers the byte to the tmux session's pane.

## Likely angles to investigate

1. **Detection condition is too strict.** macOS screenshots on the
   clipboard don't always come through as \`kind === "file"\` — the
   image data may arrive as \`kind === "string"\` with a
   \`type === "image/png"\`, or the \`items\` list might only carry
   text fallbacks. A capture-time \`console.log\` of
   \`Array.from(items).map(it => [it.kind, it.type])\` against a
   real Cmd+V will tell us which.
2. **\`\\x16\` doesn't trigger the agent CLI's clipboard read in this
   environment.** claude-code / codex listen for Ctrl-V on stdin in
   raw mode; the byte may be arriving but the CLIs may require the
   keystroke to come from a TTY in the foreground app rather than
   from another process's stdin write. tmux + the runtime's
   \`send_bytes\` path should still be a clean keystroke, but worth
   confirming with a manual \`tmux send-keys C-v\` against the same
   pane.
3. **xterm.js paste handler ran first.** The listener is registered
   with \`{ capture: true }\`, but xterm.js may also bind in capture
   phase — first-registered wins. If xterm.js \`addEventListener\`
   ran before \`RunnerTerminal\`'s init effect, our handler never
   sees the event with image data; we just see the placeholder text
   via xterm.js's own pipeline.
4. **Tauri WebKit paste-event quirks.** macOS WebKit historically
   strips clipboard \`File\` items for non-input elements; the
   xterm.js textarea is a real input but may still surface only
   text. Worth testing on a hosted WebView build to confirm.

## Environment

- OS: macOS (reported)
- Version: 0.1.4

yicheng47:
Closes #79.

## Problem

`Cmd+V` of a screenshot into a runner chat's embedded terminal landed in the agent as a "generic macOS PNG file icon, not actual content." The same paste from a host `Terminal.app` running `claude` worked fine — only Runner's embedded terminal failed.

## Why

WKWebView is in the path. When it dispatches the JS `paste` event, it materializes the image clipboard item into a JS `File` by writing the bytes to a temp file under the hood, and as a side effect mutates NSPasteboard so `public.png` now resolves to the OS-rendered icon of *that temp file* rather than the original screenshot bytes. `preventDefault()` doesn't help — the mutation happens before JS sees the event. By the time the agent CLI's `pbpaste` runs, NSPasteboard's `public.png` is the file-icon PNG.

## Fix

Read the original bytes off the `ClipboardEvent`'s `File` (still intact at that point), ship them to Rust, write to `$TMPDIR/runner-paste-<ulid>.<ext>`, then re-populate NSPasteboard with the real bytes via:

```
osascript -e 'set the clipboard to (read POSIX file "<path>" as «class PNGf»)'
```

Frontend then injects Ctrl-V (`\x16`). The agent's existing paste flow runs unchanged and shows its native `[Image x]` placeholder with the correct attachment — same UX as a host-terminal paste.

## Considered & rejected

- **`\x16`-only** (the original implementation): structurally broken across the webview boundary — NSPasteboard is no longer the clipboard the user copied.
- **Type the path into the prompt**: gets the bytes through but loses the agent's native `[Image x]` rendering. Worse UX than a host paste.
- **Clipboard-restore + `\x16`** (this PR): preserves native UX, no per-CLI custom protocol.

Full design notes in `docs/impls/0006-image-paste.md`.

## Verification

- [x] Sanity-test the osascript syntax locally: `clipboard info` reports `«class PNGf»` populated at the correct byte count after the call.
- [x] User confirmed end-to-end paste of a real screenshot now shows `[Image x]` with the actual content in the runner chat.
- [x] `pnpm tsc --noEmit`, `pnpm lint`, `cargo check`, `cargo clippy` — all clean.

## Test plan

- [ ] `Shift+Cmd+Ctrl+4` → Cmd+V in a runner chat's terminal → agent shows `[Image x]` with the real screenshot.
- [ ] `Cmd+C` a PNG file in Finder → Cmd+V in a runner chat's terminal → same `[Image x]` rendering with the file's real bytes.
- [ ] Plain-text paste still works (xterm.js default path).
- [ ] Pasting with no live session is a silent no-op.

### DIFF
--- a/docs/features/07-zoom-and-font-size.md
+++ b/docs/features/07-zoom-and-font-size.md
@@ -1,4 +1,4 @@
-# 07 — App zoom & terminal font size in General settings
+# 07 — App zoom (General) & terminal font size (new Terminal pane)
 
 > Tracking issue: [#78](https://github.com/yicheng47/runner/issues/78)
 
@@ -17,10 +17,16 @@ Two recurring asks from users on different display sizes:
    own renderer and needs its own font-size knob to stay crisp; a
    webview zoom on a canvas-rendered terminal blurs.
 
-The General pane in Settings is the right home: it already owns
-"Defaults and startup behavior" and is where users go to look first.
-Two new rows: one for **App zoom** (whole-UI scale) and one for
-**Terminal font size** (xterm-only).
+Settings split:
+
+- **App zoom** lives in the existing **General** pane — it's a
+  whole-app affordance and belongs next to other startup/defaults.
+- **Terminal font size** lives in a new dedicated **Terminal** pane
+  in the Settings sidebar. It's the first xterm-specific knob, but
+  it won't be the last: cursor style, scrollback length, copy-on-
+  select, shell preference are all natural future neighbours. Giving
+  terminal settings their own home now avoids overstuffing General
+  later.
 
 ## Scope
 
@@ -37,7 +43,13 @@ Two new rows: one for **App zoom** (whole-UI scale) and one for
   - Applied at app boot (a small effect in `App.tsx` reads the stored
     value and calls `setZoom` once on mount) and immediately on
     change.
-- **Terminal font size row** in `GeneralPane`:
+- **New `TerminalPane`** in `SettingsModal.tsx`:
+  - Add `"terminal"` to the `Pane` union and an entry to `PANES`
+    (lucide `Terminal` icon, subtitle e.g. "xterm appearance").
+    Insert between **General** and **Updates** in the sidebar.
+  - Render a `<TerminalPane />` component that hosts the terminal-
+    scoped rows; future terminal settings land here.
+- **Terminal font size row** in the new `TerminalPane`:
   - Control: a `StyledSelect` with `Small (12)`, `Default (13)`,
     `Large (15)`, `Extra large (17)`. Default `Default (13)`.
   - Persisted in localStorage at `settings.terminalFontSize`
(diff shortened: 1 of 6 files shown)


yicheng47:
Review findings:

1. Medium: non-PNG images are accepted but written to the clipboard as PNG. `src/components/RunnerTerminal.tsx:236` accepts any `image/*`, but `src-tauri/src/commands/session.rs:188` always reads the temp file as `«class PNGf»` without converting bytes. JPEG/GIF/WebP pastes can become a PNG clipboard flavor containing non-PNG bytes. Either restrict interception to `image/png` or transcode to PNG before restoring NSPasteboard.

2. Medium: pasted screenshots are left on disk. `src-tauri/src/commands/session.rs:171` writes every paste to `$TMPDIR/runner-paste-*.{ext}`, but nothing removes it after `osascript` finishes or fails. These files can contain sensitive screenshots and accumulate over time. Use `tempfile::NamedTempFile` or explicitly clean up after the clipboard restore path.

Open question: PR body says `docs/impls/0006-image-paste.md` has full design notes, but the committed file is empty.

Verification:
- `pnpm run lint` passed with the existing Fast Refresh warning in `UpdateContext.tsx`.
- `cargo check` passed.

PR #89, opened 2026-05-12
fix(session): delay verified first-prompt submit

### LINKED ISSUE #88: bug: mission start leaves the launch prompt in every slot's input — Enter never lands
## Description

When starting a mission, the composed launch prompt / persona text is delivered to **every** slot's input editor (lead + workers) but **none of the slots submit it**. Exactly one copy of the body sits in each slot's input area. The agents never start working until the human manually presses Enter in each slot.

Plan 0005 / PR #72 (`inject_paste_with_verify`) was meant to make this deterministic. The fix is in place at `src-tauri/src/session/manager.rs:1303`, but the symptom is still reproducible.

## Expected behavior

`mission_start` should:
- Compose the launch prompt for the lead and deliver it via `Router::inject_and_submit_delayed` → `inject_paste_with_verify` → paste + verify + Enter.
- Compose each non-lead's persona via `schedule_mission_first_prompt` → `inject_first_turn` → `inject_paste_with_verify`.
- Every slot's first user turn should be submitted automatically — no manual Enter required.

## Observed behavior

Across every slot (lead and workers):
- The body appears verbatim **once** in the agent's input editor.
- Enter never fires (or fires but is ignored).
- The agents sit idle until a human Enter is pressed per slot.

## Steps to reproduce

1. Start the default Build squad mission (3 slots: architect lead + impl + reviewer workers, all on claude-code).
2. Provide any non-empty mission goal.
3. Observe each slot's pane in the workspace — every slot shows the launch prompt / persona text in its input area but never submits.

Affects v0.1.6.

## Diagnosis

\"One copy in input, no Enter, across all slots\" rules out my first guess (verify rejected all 4 retries → 4 stacked pastes). The uniform behavior across independently-spawned slots also rules out a per-agent boot-ready race.

That leaves two candidates inside the verified-paste primitive:

### Candidate A — verify accepts, but `send_key(\"Enter\")` doesn't submit in claude-code

`SessionRuntime::send_key(pane, \"Enter\")` shells out to \`tmux send-keys -t=<pane> Enter\`, which sends the literal CR (`\\r`) to the pty. For most TUIs that's the standard submit. In claude-code's current TUI, possibilities:

- **Render lag**: at the moment verify accepts (count delta on the head/tail marker or the `Pasted text` placeholder), claude-code's input editor hasn't fully \"committed\" the placeholder yet — Enter lands while the editor is still in a transitional state and is dropped or treated as a no-op.
- **Multi-line state**: a long pasted body may put claude-code's composer into a state where Enter inserts a newline rather than submitting (similar to the Shift+Enter newline path we patched in xterm-side). The user has to press another key (e.g. Ctrl+Enter, Shift+Enter, or just Enter again) to actually submit.
- **Alternate screen**: claude-code may be on a transient screen (banner, model picker, trust dialog) at the verify-accept moment that consumed the Enter but didn't route it to the input editor.

This is the more likely candidate given that the body *did* land visibly (one copy).

### Candidate B — verify rejected all attempts, only the first paste landed

Even though we paste up to 4 times, only the first paste might actually reach the input editor:

- claude-code may dedupe identical adjacent bracketed pastes inside a short window, so retries 2-4 collapse into a no-op.
- `tmux load-buffer -b <name> -` then `paste-buffer -p -r -d -b <name> -t=<pane>` — the `-d` deletes the named buffer after paste, but a transient daemon hiccup could fail load-buffer on retries (would return Err from `paste` and bubble up; `inject_paste_with_verify` would also short-circuit on that Err).

This explains \"1 copy, no Enter\" if verify's three OR'd signals all miss for the first capture and the loop exhausts attempts. Less likely than A given the uniform cross-slot pattern, but possible.

## Relevant code

- \`src-tauri/src/session/manager.rs:1303\` — \`inject_paste_with_verify\` (the retry/verify loop; calls \`send_key(\"Enter\")\` on first delta-positive attempt and returns Ok)
- \`src-tauri/src/session/manager.rs:1902\` — \`FIRST_PROMPT_CONFIG\` (initial 1500ms / render 600ms / between 800ms / attempts 4)
- \`src-tauri/src/session/manager.rs:2128\` — \`paste_markers\` (head/tail marker extraction)
- \`src-tauri/src/session/manager.rs:2163\` — \`strip_ansi\` (CSI stripping before substring match)
- \`src-tauri/src/session/tmux_runtime.rs:644\` — \`send_key\` (\`tmux send-keys -t=<pane> Enter\`)
- \`src-tauri/src/session/tmux_runtime.rs:568\` — \`paste\` (\`load-buffer\` + \`paste-buffer -p -r -d\`)
- \`src-tauri/src/router/mod.rs:456\` — \`inject_and_submit_delayed\` (lead launch prompt routing)
- \`src-tauri/src/session/manager.rs:2028\` — \`inject_first_turn\` (per-slot persona injection)
- \`docs/impls/0005-first-prompt-readback.md\` — original plan

## Environment

- OS: macOS
- Version: v0.1.6
- Runtime: claude-code

## Investigation gap — observability

Runner.app's stdout/stderr are wired to \`/dev/null\` under Finder/launchd, so the verify loop's existing \`eprintln!\` lines (\`runner: first-prompt for {session_id}: paste not visible after N attempts\` and the per-attempt capture failures) are invisible in Console.app and \`log show\`. Confirming Candidate A vs B requires either:

1. **Terminal relaunch**: kill Runner.app and run \`/Applications/Runner.app/Contents/MacOS/runner 2>&1 | tee /tmp/runner.log\` to capture stderr live for one reproduction.
2. **File logging follow-up**: add \`\$APPDATA/runner/logs/runner.log\` so subsequent regressions are observable without the dance. Worth doing regardless.

## Suspected fix directions

- **If A**: insert a small post-verify-accept settle delay (e.g. 100-200ms) before `send_key(\"Enter\")` so claude-code's input editor has time to commit the placeholder; OR send a sequence (e.g. End → Enter, or Down → Enter) that nudges the composer out of any newline-insert state; OR replace `send_key(\"Enter\")` with explicit `send_bytes(b\"\\r\")` and verify the byte arrives in raw mode.
- **If B**: add a per-attempt log line so we can tell which signal (head/tail/placeholder) is delivering the match, or whether none are; possibly capture the actual stripped bytes when verify rejects so we can see what claude-code rendered.
- **Operational**: ship file logging now so we stop flying blind on injection bugs.

yicheng47:
## Summary
- add a 120ms submit_wait after first-prompt paste verification before sending Enter
- keep tests synchronous with zero submit_wait under cfg(test)

Closes #88

## Validation
- cargo test -p runner first_prompt

### DIFF
--- a/src-tauri/src/session/manager.rs
+++ b/src-tauri/src/session/manager.rs
@@ -1292,9 +1292,9 @@ impl SessionManager {
     /// up to `max_attempts` rounds of paste → sleep `render_wait` →
     /// capture → if any of head/tail-marker delta or (body ≥
     /// `PLACEHOLDER_MIN_BODY_LEN`) placeholder delta ≥ 1 vs the
-    /// baseline, send Enter and return. Otherwise sleep
-    /// `between_attempts` and retry. If no attempt verifies, return
-    /// Err — caller logs.
+    /// baseline, wait `submit_wait`, send Enter and return. Otherwise
+    /// sleep `between_attempts` and retry. If no attempt verifies,
+    /// return Err — caller logs.
     ///
     /// `before` capture failures fall through with zero baselines
     /// (alternative is to abort, which regresses every transient
@@ -1371,6 +1371,9 @@ impl SessionManager {
                 && count_substr(&after_stripped, b"Pasted text") > before_placeholder_count;
 
             if head_delta_pos || tail_delta_pos || placeholder_delta_pos {
+                if !config.submit_wait.is_zero() {
+                    std::thread::sleep(config.submit_wait);
+                }
                 return self
                     .runtime
                     .send_key(&rt_session, "Enter")
@@ -1884,16 +1887,19 @@ fn capture_cwd(explicit: Option<String>) -> Option<String> {
 /// Tunables for the first-prompt readback loop. Production uses a
 /// short initial wait (so a fast spawn doesn't sit idle), modest
 /// per-attempt render wait (let tmux + the agent TUI commit the
-/// paste before we capture-pane), and a small max_attempts. `cfg(test)`
-/// zeros every duration so unit tests stay synchronous; the count
-/// stays at 4 so retry/give-up paths still exercise their branches.
+/// paste before we capture-pane), a submit wait after verification
+/// (let the editor leave bracketed-paste mode before Enter), and a
+/// small max_attempts. `cfg(test)` zeros every duration so unit tests
+/// stay synchronous; the count stays at 4 so retry/give-up paths still
+/// exercise their branches.
 ///
 /// See `docs/impls/0005-first-prompt-readback.md` for the rationale
-/// behind the specific numbers (best case 2100ms, worst case 7100ms).
+/// behind the specific numbers (best case 2220ms, give-up 7100ms).
 #[derive(Debug, Clone, Copy)]
 pub(crate) struct FirstPromptConfig {
     pub initial_wait: std::time::Duration,
     pub render_wait: std::time::Duration,
+    pub submit_wait: std::time::Duration,
     pub between_attempts: std::time::Duration,
     pub max_attempts: usize,
 }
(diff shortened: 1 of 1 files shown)


yicheng47:
Closing in favor of a structural fix tracked in #88: deliver the first user turn via the agent CLI's positional `[PROMPT]` argv at spawn time, instead of post-spawn paste + Enter. Empirically verified both `claude "hello"` and `codex "hello"` (codex-cli 0.130.0) accept the positional — the codebase's earlier "approval dialog swallows the positional" caveat in `router/runtime.rs:13-16` is stale.

Spawn-time argv delivery eliminates the paste → render-wait → capture → verify → Enter race entirely. Plan doc + impl on a new branch; will link the PR here when it's up.

The 120ms `submit_wait` this PR introduces would be obsolete on day one of the new path.

PR #90, opened 2026-05-12
feat(session): deliver first user turn via spawn-time positional argv

### LINKED ISSUE #88: bug: mission start leaves the launch prompt in every slot's input — Enter never lands

yicheng47:
## Summary

- Closes #88 by moving first-user-turn delivery (lead launch prompt, per-slot persona, direct-chat persona) from the post-spawn paste path to the agent CLI's positional `[PROMPT]` argv at process spawn.
- The agent reads its argv during init — before the TUI binds raw input, before claude-code's trust-folder dialog, before the input editor exists — so the race that left the message buffered without an Enter is eliminated for the dominant case.
- Empirically verified that both `claude "hello"` and `codex "hello"` accept the positional in current CLI versions; the codebase's stale comment about codex's positional being swallowed by an approval dialog has been corrected.

## Architecture

Plan: [`docs/impls/0007-spawn-time-prompt-delivery.md`](https://github.com/yicheng47/runner/blob/feat/spawn-time-prompt-delivery/docs/impls/0007-spawn-time-prompt-delivery.md)

| Layer | Before | After |
|---|---|---|
| **Composition** | `router::handlers::mission_goal` builds the launch prompt post-spawn from bus state. Worker preamble composed inside `manager::schedule_mission_first_prompt`. Direct-chat persona composed inside `manager::schedule_direct_first_prompt`. | All three composers live in `router::prompt` (`compose_launch_prompt`, `compose_worker_first_turn`, `compose_direct_first_turn`). Both delivery paths (argv + paste fallback) share byte-identical bodies. |
| **Delivery — fresh spawn** | `inject_paste_with_verify` → paste → 600ms render-wait → capture-pane → marker delta → send Enter. Up to 4 attempts. Race-prone against TUI boot. | Body lands as the trailing positional argv at process spawn. No waits, no readback, no Enter call. |
| **Delivery — resume** | Same paste-verify path. | Unchanged — agent CLIs restore prior context via their own session resume. The rare resume-fresh-fallback (`Router::fire_lead_launch_prompt`) keeps paste-verify. |
| **`mission_goal` handler** | Composes the launch prompt and calls `inject_and_submit_delayed`. | No-op for delivery. Stays subscribed for UI consumers and future signal-routing extensions. |
| **ARG_MAX guard** | Not applicable. | `FIRST_TURN_ARGV_MAX_BYTES = 32 KB` defensive cap; bodies over fall back to the existing paste path with no behavioral regression. |

## Key changes

- `router/runtime.rs` — new `first_turn_argv(runtime, body)` helper; `trailing_runtime_args` grows a `first_turn` parameter that rides on the same resume-suppression guard as the existing `system_prompt` argv.
- `router/prompt.rs` — worker preamble + helper composers moved here from `manager.rs`.
- `router/handlers.rs` — `mission_goal` no-ops the launch-prompt injection. Vestigial `LEAD_LAUNCH_PROMPT_DELAY` dropped.
- `session/manager.rs` — `apply_runtime_args` returns `delivered_via_argv`. `spawn` / `spawn_direct` take a `first_turn: Option<String>` and skip `schedule_*_first_prompt` when argv took over.
- `commands/mission.rs` — `mission_start` and `mission_reset` compose per-slot bodies before the spawn loop and pass them in.
- `commands/session.rs` — `session_start_direct` composes the direct persona and passes it in.

## Tests

- Existing: 212 → 216 lib + 44 workspace.
- Two router tests (`mission_goal_injects_composed_prompt_to_lead`, `lead_launch_prompt_routes_through_verified_paste`) collapsed into a single guard that the handler is now silent for the lead. `fresh_mission_start_does_not_call_reconstruct…` keeps the reconstruct-skip invariant, drops the launch-prompt assertion.
- New: 5 router/runtime tests (`first_turn_argv_*`, fresh-spawn argv coverage, resume suppression) + 2 manager tests (`direct_chat_persona_lands_as_trailing_positional_when_argv_delivered`, `mission_spawn_worker_preamble_lands_as_trailing_positional_when_argv_delivered`).
- `cargo test --workspace` — all green.
- `cargo clippy -- -D warnings` — clean.

## Test plan

- [ ] Start the default Build squad mission in a never-trusted directory. Accept claude-code's trust dialog. The launch prompt should be delivered immediately — no manual Enter per slot.
- [ ] Start a mission with a long composed prompt (≥2 KB). All slots should pick up their first turn at boot.
- [ ] Start a direct chat with a non-trivial persona. Persona lands as the first user turn, no manual Enter.
- [ ] Resume a previously-stopped mission with intact claude conversation. Agent restores prior context via its own session resume — no re-injection.
- [ ] Resume a mission where claude's conversation file is missing (fresh-fallback). `fire_lead_launch_prompt` paste-verify path still works.

### DIFF
--- a/docs/impls/0007-spawn-time-prompt-delivery.md
+++ b/docs/impls/0007-spawn-time-prompt-delivery.md
@@ -210,23 +210,31 @@ post-spawn. After this plan:
 - On resume of a non-lead, the agent's own session resume restores
   the persona context; no re-injection needed for either runtime.
 
-### ARG_MAX guard
+### Persist-time validation (no fallback)
 
-macOS `ARG_MAX` is ~256KB. Mission prompts are typically 1-5KB.
-Add a defensive cap:
+`first_turn_argv` is now the **only** path for the first user turn
+on a fresh spawn. There is no paste fallback for oversize bodies —
+instead, persistence-layer validation rejects inputs that would
+push the composed body past the runtime's argv slot.
 
-```rust
-const SPAWN_ARGV_PROMPT_MAX_BYTES: usize = 32 * 1024; // 32KB
+Two caps, both surfaced in the corresponding `commands::*` module:
 
-if first_turn.len() > SPAWN_ARGV_PROMPT_MAX_BYTES {
-    log + fall back to post-spawn paste-verify for this slot
-}
-```
+| Constant | Site | Value |
+|---|---|---|
+| `MAX_SYSTEM_PROMPT_BYTES` | `commands::runner` | 16 KB |
+| `MAX_MISSION_GOAL_BYTES` | `commands::mission` (reused by `commands::crew::validate_crew_goal`) | 8 KB |
 
-32KB leaves an 8x safety margin against the OS cap and trips well
-before any realistic mission prompt. The fallback path is the
-existing `schedule_mission_first_prompt` / `inject_paste_with_verify`
-machinery — unchanged.
+Sized so the composed launch prompt — preamble (~1 KB) +
+`system_prompt` (≤16 KB) + goal (≤8 KB) + roster (~2 KB) +
+coordination (~1 KB) — stays well under
+`router::runtime::FIRST_TURN_ARGV_MAX_BYTES` (32 KB, defense-in-
+depth). The worker preamble and direct-chat persona use only a
+subset of those fields and are bounded by the same caps.
+
+`first_turn_argv` carries a `debug_assert!` on the
+`FIRST_TURN_ARGV_MAX_BYTES` ceiling so a future drift in the
+composition logic that bypasses validation surfaces in tests
+rather than silently truncating at spawn.
 
 ### Trust-folder dialog (claude-code)
 
@@ -299,9 +307,14 @@ them forward.
 ## Risks
 
 - **ARG_MAX edge cases on weird systems.** macOS is fine. Linux
-  varies (per `getconf ARG_MAX`); typical 2MB. The 32KB cap is
-  defensive. Embedded environments with smaller limits would trip
-  the fallback path — acceptable.
+  varies (per `getconf ARG_MAX`); typical 2MB. The 32 KB
+  defense-in-depth ceiling in `first_turn_argv` is the secondary
+  guard; the primary guard is the persistence-layer cap on the
+  composing fields (`system_prompt` ≤ 16 KB, `goal` ≤ 8 KB), which
+  prevents the body from approaching the ceiling in the first
+  place. Embedded environments with `ARG_MAX` below 32 KB would
+  need a smaller persist-time cap — punted until a real platform
+  surface asks for it.
 
 - **Shell-quoting bugs.** The existing
   `render_launch_script` quoter handles single quotes; multi-line
@@ -359,9 +372,14 @@ them forward.
 4. **`direct_chat_passes_persona_via_argv`** (session.rs) —
    `session_start_direct` calls spawn with `first_turn = persona`.
 
-5. **`spawn_argv_too_long_falls_back_to_paste`** (mission.rs or
-   manager.rs) — a body > 32KB triggers the fallback path and
-   `inject_paste_with_verify` is called instead.
+5. **`create_rejects_system_prompt_over_cap`** /
+   **`update_rejects_system_prompt_over_cap`** (commands::runner) —
+   `system_prompt` over `MAX_SYSTEM_PROMPT_BYTES` is rejected at
+   create + update; no DB write happens.
+   **`create_rejects_goal_over_cap`** /
+   **`update_rejects_goal_over_cap`** (commands::crew) — same for
+   `crew.goal`. **`start_rejects_goal_override_over_cap`**
+   (commands::mission) — same for per-mission goal_override.
 
 6. **`mission_goal_handler_no_op_when_prompt_spawn_delivered`**
    (router) — bus replay of `mission_goal` against a fresh mission
--- a/docs/impls/0007-spawn-time-prompt-delivery.md
+++ b/docs/impls/0007-spawn-time-prompt-delivery.md
@@ -1,0 +1,410 @@
+# Spawn-time first-prompt delivery
+
+> Replaces the post-spawn paste → capture-verify → Enter dance for the
+> *first user turn* (mission lead launch prompt + per-slot personas +
+> direct-chat personas) with a positional CLI argument passed at
+> process spawn. Eliminates the readback race entirely for the boot
+> path. Targets the next patch release.
+
+## Why
+
+Issue #88 (also referenced in #50, partially mitigated by plan 0005):
+after `mission_start`, the composed launch prompt / persona text
+sometimes lands in the agent's input editor but is never submitted —
+exactly **one** copy of the body sits in every slot's input, no Enter
+fires, all slots stuck. The user has to press Enter in each slot to
+unstick the mission.
+
+Plan 0005 / PR #72 (`inject_paste_with_verify`) was meant to make the
+paste→Enter handoff deterministic via a capture-pane readback loop.
+It's strictly better than the prior `FIRST_PROMPT_DELAY = 2500ms`
+blind wait, but it still has gaps:
+
+- **Render-lag at verify-accept.** Even when the marker delta fires
+  on attempt 1, claude-code's input editor may not be in a "ready to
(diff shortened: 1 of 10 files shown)


yicheng47:
Review findings:

1. Medium: the lead launch prompt is dropped when it exceeds the argv cap. `first_turn_argv` returns empty for bodies over `FIRST_TURN_ARGV_MAX_BYTES` (`src-tauri/src/router/runtime.rs:468`), but the lead spawn still calls `schedule_mission_first_prompt` with `suppress_lead_preamble = true` (`src-tauri/src/session/manager.rs:544`), and that function returns before checking `delivered_via_argv` (`src-tauri/src/session/manager.rs:2021`). Since `mission_goal` is now a no-op (`src-tauri/src/router/handlers.rs:39`), there is no paste fallback for a large lead launch prompt on `mission_start` or `mission_reset`. Non-lead and direct-chat fallbacks still work, but the lead comes up with no goal/roster/coordination context exactly in the documented fallback case. Pass the composed lead `first_turn` into the fallback path, or otherwise only suppress the lead paste path when argv delivery actually happened.

Verification:
- `cargo test --workspace` passed: 216 lib tests, 11 CLI tests, 13 roundtrip tests, 20 runner-core tests; 6 ignored.

yicheng47:
Review findings:

1. Medium: the new persist-time caps still do not guarantee the lead argv body fits. `first_turn_argv` still drops the first turn in release when the composed body exceeds `FIRST_TURN_ARGV_MAX_BYTES` (`src-tauri/src/router/runtime.rs:480`), and now debug builds panic first via `debug_assert!` (`src-tauri/src/router/runtime.rs:474`). The fix caps `system_prompt` and mission/crew goal, but the lead body also includes `crew_name`, the lead display name, and every roster entry (`slot_handle` + runner `display_name`) from `mission_start` (`src-tauri/src/commands/mission.rs:545`) and `compose_launch_prompt` (`src-tauri/src/router/prompt.rs:107`, `src-tauri/src/router/prompt.rs:137`). Those fields are only checked for non-empty in the create/update paths, not length-capped (`src-tauri/src/commands/crew.rs:191`, `src-tauri/src/commands/runner.rs:311`, `src-tauri/src/commands/slot.rs:213`). A single very long display name/crew name or a large roster can still push the body past 32 KB, so the launch prompt is again absent in production and can panic in dev. Add a final composed-body length check before spawning, or cap all fields/roster size that contribute to the launch prompt.

Verification:
- `cargo test --workspace` passed: 217 lib tests, 11 CLI tests, 13 roundtrip tests, 20 runner-core tests; 6 ignored.

PR #151, opened 2026-05-17
wip(terminal): resize-aligned attach snapshot (#150)

### LINKED ISSUE #150: bug: terminal history drifts after cold-restart + mid-session window resize (claude-code only)
## Description

After quitting and relaunching the app, claude-code panes show **drifted/stacked terminal history** when the user resizes the window mid-session. Each line in the scrollback has progressively-shifting indentation, with the visible right-edge of the original render still baked in at mid-screen.

Only affects **claude-code**. Codex and plain shell sessions render correctly through the same reattach path.

## Reproduction

1. Open a mission with a claude-code lead. Let it accumulate some history.
2. **Quit the app** (Cmd+Q). Wait for full process exit.
3. **Relaunch the app.** Open the same mission. Switch to the `@architect` (or any claude-code) tab.
4. Drag the window edge to resize it.
5. Scroll up in the terminal pane → broken/drifted layout in the historical view.

## Why claude-code only

claude-code's TUI (Ink-based) emits **absolute cursor-positioning escapes** (`ESC[<row>;<col>H`) for every cell. When replayed into an xterm at a different cols than the original render, cells land at wrong absolute coordinates — text scatters with the rendering-box edges baked in at the wrong place. Codex / Ink-less TUIs use relative positioning and reflow correctly.

## Why this isn't covered by #148 / #149

PRs #148 (docs) + #149 (code) landed the 0009 fix: xterm enters alt-screen at reattach so live redraws don't stack in main-screen scrollback. That covers the *running-session* path.

The remaining gap is **the moment of reattach itself**: tmux's pane still has whatever cols it was at when the app quit (e.g. 200). xterm mounts at the constructor default (cols=80). We replay 200-cols-wide bytes into an 80-cols xterm, baking the wrap mismatch into xterm's alt-screen buffer. When the user then resizes the window, claude-code's *diff-based* SIGWINCH redraw only updates the cells it knows have changed — the cells we mis-painted at startup stay broken forever.

## Investigation history

- `docs/impls/0009-terminal-alt-screen-reattach.md` — the live-session fix that landed in #149.
- `docs/impls/0010-terminal-replay-cols-alignment.md` — shelved earlier. The plan as drafted resized xterm to match tmux's pane cols, which fixes the *current page* drift but not previous-page drift (claude-code never re-emits its internal scroll history).

## Better approach (per follow-up research)

Looking at how VS Code does persistent terminal restore via `SerializeAddon` confirms the underlying constraint: the serialized state of an xterm-like buffer is **size-dependent**, so cols at serialize-time must match cols at replay-time. VS Code solves this by having the frontend send its cols to the server first, the server resizes its headless xterm + SIGWINCHes the agent, *then* serializes.

For us, the equivalent is to flip 0010's direction: instead of telling xterm what cols the snapshot was captured at, have the frontend tell the backend its container cols *first*, then capture. Concretely:

1. Frontend mounts at cols=80, subscribes + buffers live events (current behaviour).
2. Activation effect: `fit.fit()` → xterm at container cols (say 200).
3. New API: `session_attach_snapshot(sessionId, cols, rows)` that atomically (a) resizes tmux to the requested cols (SIGWINCH → claude-code redraws at the new cols), (b) re-runs `capture_replay_bytes` at the post-resize size, (c) returns the fresh bytes.
4. Frontend writes the returned bytes — xterm cols ≡ tmux cols ≡ claude-code's draw cols, by construction. No mismatch possible.
5. Flush buffered live events as today.

This also obviates the 'previous-page drift' worry that originally shelved 0010 — we never replay claude-code's old internal-scroll history; we always replay the current screen, which is always at the right cols. Users wanting to see chat history go through claude-code's own UI (same as iTerm2).

## Files likely touched

- `src-tauri/src/session/manager.rs` — new `output_snapshot` variant (or a separate `attach_snapshot` method) that takes target cols/rows.
- `src-tauri/src/session/tmux_runtime.rs` — `capture_replay_bytes` already exists; the new path resizes first, then captures.
- `src-tauri/src/commands/session.rs` — IPC surface.
- `src/lib/api.ts` + `src/components/RunnerTerminal.tsx` — call the new API from the activation effect instead of the current mount-time `outputSnapshot`.

## Acceptance

- Quit + relaunch + window resize: claude-code's pane renders clean, no drifted historical lines, no broken vertical "right-edge" of the original render visible.
- Codex / plain shell / non-alt-screen sessions: unaffected (regression test).
- Live-session behaviour from #149: still works (no regressions on tab-switch / mid-session resize).

yicheng47:
> **Draft / WIP.** Improves the cold-restart symptom but doesn't fully fix #150. Opening for visibility on the work-in-progress shape; not ready to merge.

## Summary

Implements the resize-then-capture attach flow proposed in #150:

- Frontend tells the backend its post-\`fit.fit()\` container cols/rows.
- Backend resizes tmux → waits 200 ms for the agent's SIGWINCH redraw to flow through pipe-pane → \`capture-pane\` returns fresh cells at the matched dims (with the 0009 alt-screen prepend).
- Backend reads \`output_seq\` immediately after the capture as a high-water-mark.
- Frontend writes the bytes into an xterm sized to the same cols/rows, then dedups buffered live events against \`as_of_seq\` to avoid double-painting.

## Backend

- New \`SessionRuntime::resize_and_capture\` trait method; tmux impl runs \`resize-window\` → 200 ms sleep → \`capture_replay_bytes\`.
- \`SessionManager::attach_snapshot\` orchestrates the call and reads the high-water-mark.
- \`AttachSnapshot { data, as_of_seq }\` struct + \`session_attach_snapshot\` Tauri command.
- \`InertRuntime\` / \`FakeRuntime\` test stubs.

## Frontend

- Mount effect now only subscribes + buffers live events into \`pendingLiveRef\`; the snapshot fetch moves to the activation effect so it runs *after* \`fit.fit()\` and we know the container cols.
- Activation effect (first-activation-only) calls \`attachSnapshot(cols, rows)\`, writes the returned bytes, sets \`lastWrittenSeqRef = as_of_seq\`, and drains \`pendingLiveRef\` skipping seqs already baked into the capture.
- \`clearVersion\` now also marks replay-done + drops buffered live so the resume flow's clean-slate intent isn't clobbered.

## Status

Manual smoke shows noticeable improvement on the cold-restart + window-resize reproducer but **doesn't fully eliminate** the misalignment. Suspect candidates for the residual gap:

- 200 ms post-resize wait may be insufficient when claude-code is mid-render; capture sees a half-redrawn frame and bakes the wrong cells. → try FIFO-quiet detection instead of a fixed sleep, or extend the wait window.
- claude-code's diff-based redraw might leave pre-resize cells untouched in rows it doesn't consider "dirty" → may need a \`clear + SIGWINCH\` to force a full repaint before capture.
- 0009's \`ESC[?1049h\` prepend may interact with state already on alt-screen unexpectedly.

Holding the branch as draft while we investigate.

## Checks

- [x] \`cargo check --all-targets\` clean
- [x] \`cargo test --lib session::manager\` (28 passing)
- [x] \`pnpm lint\` no new warnings
- [ ] Manual cold-restart verification — residual misalignment, needs more iteration

### DIFF
--- a/src-tauri/src/commands/session.rs
+++ b/src-tauri/src/commands/session.rs
@@ -20,7 +20,7 @@ use crate::{
     commands::runner,
     error::{Error, Result},
     model::{Session, SessionStatus, Timestamp},
-    session::manager::{OutputEvent, SessionEvents, SpawnedSession, TauriSessionEvents},
+    session::manager::{AttachSnapshot, OutputEvent, SessionEvents, SpawnedSession, TauriSessionEvents},
     AppState,
 };
 
@@ -140,6 +140,22 @@ pub async fn session_output_snapshot(
     session_id: String,
 ) -> Result<Vec<OutputEvent>> {
     Ok(state.sessions.output_snapshot(&session_id))
+}
+
+/// Resize the pane to xterm's container dims and return a fresh
+/// alt-screen-aware capture. Replaces `session_output_snapshot` in
+/// the frontend's mount-time attach flow — the cols-matched capture
+/// avoids the absolute-positioning misalignment claude-code's Ink
+/// renderer produces when the replay cols don't match the original
+/// draw cols (issue #150).
+#[tauri::command]
+pub async fn session_attach_snapshot(
+    state: State<'_, AppState>,
+    session_id: String,
+    cols: u16,
+    rows: u16,
+) -> Result<AttachSnapshot> {
+    state.sessions.attach_snapshot(&session_id, cols, rows)
 }
 
 /// Restore NSPasteboard for a PNG paste that came in through the
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -282,6 +282,7 @@ pub fn run() {
             commands::session::session_kill,
             commands::session::session_resize,
             commands::session::session_output_snapshot,
+            commands::session::session_attach_snapshot,
             commands::session::session_paste_image,
             commands::session::session_start_direct,
         ])
--- a/src-tauri/src/session/manager.rs
+++ b/src-tauri/src/session/manager.rs
@@ -256,6 +256,23 @@ pub struct OutputEvent {
     pub data: String,
 }
 
+/// Resize-aligned attach snapshot returned by `attach_snapshot`. The
+/// frontend writes `data` into an xterm sized to the cols/rows it asked
+/// for, then dedups buffered live events against `as_of_seq` (drop
+/// `seq <= as_of_seq`, replay `seq > as_of_seq`). Implements the fix
+/// for issue #150 (cold-restart cols-mismatch drift).
+#[derive(Debug, Clone, Serialize)]
+pub struct AttachSnapshot {
+    /// Base64-encoded capture-pane bytes at the requested dims, with
+    /// alt-screen-enter prepended when applicable (see
+    /// `capture_replay_bytes` and docs/impls/0009).
+    pub data: String,
+    /// `output_seq` value at the moment of capture. Live events with
+    /// `seq <= as_of_seq` are already reflected in `data` and should
+    /// not be replayed by the consumer.
+    pub as_of_seq: u64,
+}
+
 #[derive(Debug, Clone, Serialize)]
 pub struct ExitEvent {
     pub session_id: String,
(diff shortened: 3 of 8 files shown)


yicheng47:
Manual repro on the current branch: every window resize **still produces additional terminal history** — same shape as the original stacking bug, content piles up rather than the alt-screen redraw replacing in place. Even with 0009's `ESC[?1049h` prepend at attach and this PR's resize-then-capture flow, mid-session resizes keep appending.

Possible cause hypotheses for the iteration:

- xterm slips back out of alt-screen between attach and the user-driven resize (e.g. an escape in the live stream toggles the mode), so claude-code's resize redraw lands on main-screen and scrolls into scrollback like before 0009.
- The activation effect's `api.session.resize` runs every time the active tab re-activates, not just once — each call SIGWINCHes the agent and triggers a full repaint that we may be writing on top of stale alt-screen cells.
- The 200 ms post-resize wait in `resize_and_capture` lets some of claude-code's redraw bytes through pipe-pane *before* we capture, so they get recorded as Stream events with seq > `as_of_seq`, then the frontend writes them again on top of the captured cells.

Next debug step: add JS-side `plugin-log` traces around the activation/resize path on a diagnostic build, then reproduce + inspect `runner.log` to see which of the above is firing.

yicheng47:
Closing — the resize-aligned attach snapshot approach didn't fully fix #150 (claude-code's diff-based redraw defeats the resize-then-capture flow; force-alt-screen as a workaround broke scroll behavior for non-TUI sessions). Reopening #150 for a fresh look.

The diagnostic plumbing + log path split + resize dedupe that came out of the investigation are landing separately on `chore/dev-log-split` since they're independently useful.

PR #154, opened 2026-05-18
fix: stabilize terminal attach and resize replay

### LINKED ISSUE #141: bug: terminal scrollback paging feels choppy / jumpy
## Description

Scrolling back through a runner's terminal scrollback (e.g. paging up through a long claude-code TUI session in the mission workspace) feels choppy — each wheel tick lands as a hard pixel jump rather than an interpolated scroll, and the per-tick step is small enough that paging through a 6h mission's output is visibly stuttery.

## Expected behavior

Scrollback paging should feel smooth, matching iTerm2 / Terminal.app / Warp. Each wheel event should interpolate over ~120-150ms, and per-tick distance should be large enough that paging through a few thousand lines doesn't feel like clicking through a slideshow.

## Steps to reproduce

1. Start a mission with a long-running claude-code lead (anything that produces a few thousand lines of TUI output — a /loop, a large refactor, an editing session).
2. Click into the lead's tab in the mission workspace.
3. Scroll up through the scrollback (trackpad two-finger drag or mouse wheel).
4. Observe: each wheel tick is a discrete pixel jump, no interpolation; paging through deep scrollback is choppy.

## Relevant code

- `src/components/RunnerTerminal.tsx:140-170` — `new Terminal({…})` config. Currently sets `cols`, `rows`, `theme`, `fontFamily`, `fontSize`, `cursorBlink`, `cursorStyle`, `scrollback`, `allowProposedApi`. **Does not set** `smoothScrollDuration`, `scrollSensitivity`, or `fastScrollSensitivity`.
- `src/components/RunnerTerminal.tsx:165-170` — renderer setup. Tries `WebglAddon` and silently falls through to the DOM renderer on failure. No `@xterm/addon-canvas` in between. WebGL is reliable on macOS but a load failure puts users on the slowest possible renderer with no warning.
- `src/lib/settings.ts:50-52` — `TERMINAL_SCROLLBACK_OPTIONS` is `[1000, 5000, 10000, 50000]`, default `10000`. With 4 runners at 10k each that's 40k cached lines — fine for WebGL but worth noting when characterizing perceived choppiness.
- `src/pages/RunnerChat.tsx` — same pattern, missing the same config knobs.

## Suggested fix

In both `RunnerTerminal.tsx` and `RunnerChat.tsx`'s `new Terminal({…})` call:

```ts
new Terminal({
  // existing fields…
  smoothScrollDuration: 125,   // interpolate wheel scrolls; matches Warp's feel
  scrollSensitivity: 3,        // larger per-tick distance
  fastScrollSensitivity: 8,    // shift-scroll multiplier
})
```

Optional second pass: add `@xterm/addon-canvas` and load it as the middle-tier fallback so a WebGL init failure doesn't silently drop users to the DOM renderer:

```ts
try {
  term.loadAddon(new WebglAddon());
} catch {
  try { term.loadAddon(new CanvasAddon()); } catch { /* DOM */ }
}
```

Worth exposing `smoothScrollDuration` as a setting too (Settings → Terminal) for users who want instant jumps — but a sensible default is the priority.

## Environment

- OS: macOS
- Version: 0.1.13

### LINKED ISSUE #150: bug: terminal history drifts after cold-restart + mid-session window resize (claude-code only)

yicheng47:
## Summary
- Replay attach snapshots after the frontend fits xterm to a real visible grid.
- Normalize tmux capture output from LF to CRLF before writing it into xterm.
- Keep inactive terminal panes layout-visible so they stream scrollback at real dimensions.
- Debounce/no-op resize pushes to avoid accumulating intermediate Claude redraws.

## Root Cause
- Claude Code 2.1.143 uses tmux main-screen mode, so full `capture-pane` history contains resize redraw artifacts. Replaying that polluted history caused stacked output after restart/resize.
- `tmux capture-pane -p` emits bare LF rows; xterm replay needs CRLF-style row transitions, otherwise lines staircase sideways.
- Hidden panes mounted without measurable layout, so early replay/input geometry could bake in an 80-col or stale grid.

## Validation
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm exec tsc --noEmit`
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run lint`
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run build`
- `cargo test -p runner --lib`
- `git diff --check`

Notes: frontend checks pass with the existing Node engine warning in this shell (`18.20.8` vs project `>=22.12.0`), the existing Fast Refresh lint warning in `src/contexts/UpdateContext.tsx`, and the existing Vite chunk-size warning.

Fixes #150
Fixes #141

### DIFF
--- a/src-tauri/src/commands/session.rs
+++ b/src-tauri/src/commands/session.rs
@@ -20,7 +20,7 @@ use crate::{
     commands::runner,
     error::{Error, Result},
     model::{Session, SessionStatus, Timestamp},
-    session::manager::{OutputEvent, SessionEvents, SpawnedSession, TauriSessionEvents},
+    session::manager::{OutputSnapshot, SessionEvents, SpawnedSession, TauriSessionEvents},
     AppState,
 };
 
@@ -138,8 +138,18 @@ pub async fn session_resize(
 pub async fn session_output_snapshot(
     state: State<'_, AppState>,
     session_id: String,
-) -> Result<Vec<OutputEvent>> {
+) -> Result<OutputSnapshot> {
     Ok(state.sessions.output_snapshot(&session_id))
+}
+
+#[tauri::command]
+pub async fn session_attach_snapshot(
+    state: State<'_, AppState>,
+    session_id: String,
+    cols: u16,
+    rows: u16,
+) -> Result<OutputSnapshot> {
+    state.sessions.attach_snapshot(&session_id, cols, rows)
 }
 
 /// Restore NSPasteboard for a PNG paste that came in through the
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -289,6 +289,7 @@ pub fn run() {
             commands::session::session_kill,
             commands::session::session_resize,
             commands::session::session_output_snapshot,
+            commands::session::session_attach_snapshot,
             commands::session::session_paste_image,
             commands::session::session_start_direct,
         ])
--- a/src-tauri/src/session/manager.rs
+++ b/src-tauri/src/session/manager.rs
@@ -42,6 +42,7 @@ use crate::session::runtime::{
 };
 
 const MAX_OUTPUT_BUFFER_CHUNKS: usize = 4096;
+const ATTACH_SNAPSHOT_REDRAW_GRACE: Duration = Duration::from_millis(120);
 
 /// Inputs the forwarder consumer needs to translate a
 /// `RuntimeOutput::StatusTransition` into a real `runner_status`
@@ -254,6 +255,16 @@ pub struct OutputEvent {
     pub seq: u64,
     /// Base64-encoded raw bytes read from the PTY.
     pub data: String,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct OutputSnapshot {
+    pub events: Vec<OutputEvent>,
+    /// Highest per-session output seq the frontend should consider
+    /// covered by this snapshot. Live events with seq <= this value
+    /// were already recorded before snapshot capture began and can be
+    /// dropped after the replay bytes are written.
+    pub last_seq: u64,
 }
 
 #[derive(Debug, Clone, Serialize)]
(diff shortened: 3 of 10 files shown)


yicheng47:
Closing — the snapshot-replay-on-resize approach didn't actually solve the stack. Reverting and rethinking.

PR #155, opened 2026-05-18
docs(impls): 0011 — PTY host sidecar runtime; retire shelved 0010

yicheng47:
## Summary

Lands the impl plan for replacing tmux with a Runner-owned PTY host sidecar — the long-term fix for the terminal stacking class of bugs that the dropped PR #154 made visible (and failed to repair from inside tmux's text-reconstruction model).

- **Architecture**: `Tauri ↔ runner-pty-host sidecar (Unix socket) ↔ portable-pty ↔ agent CLI`. The host detaches from the Tauri main process at boot (`setsid` + double-fork on Unix; `posix_spawn` + `POSIX_SPAWN_SETSID` fallback for macOS hardened-runtime). Single-instance via `fs2` lockfile; bundled via `tauri.conf.json` `externalBin`.
- **Persistence guarantees**: agents survive webview reload (Cmd+R), Tauri app crash, and force-quit. Only PTY host process death terminates them. Documented as an explicit table in the doc so the contract is clear.
- **Resize-stack fix**: spelled out in two layers. Live path emits a scrollback-clear to xterm subscribers before forwarding SIGWINCH. Snapshot returns "tape since last clear" rather than the unbounded raw tape. A headless-emulator optimization (`vt100` / `alacritty_terminal`) is named but deferred behind a stable wire shape.
- **Phasing**: 9 steps from protocol definition through tmux removal. Direct chats get reattach survivability first; mission survivability deferred until router/event-bus ownership question is answered.

Closes PR #154's investigation thread by committing to the architecture that makes the original problem solvable.

## Files

- **New**: `docs/impls/0011-pty-host-terminal-runtime.md` (551 lines).
- **Deleted**: `docs/impls/0010-terminal-replay-cols-alignment.md` — explicitly shelved as "investigated, not implemented" and superseded wholesale.

## Test plan

- [ ] Plan reads coherently end-to-end
- [ ] No code changes — N/A for tsc/lint/cargo


yicheng47:
Review findings:

1. High: the resize clear strategy would erase normal shell scrollback. `docs/impls/0011-pty-host-terminal-runtime.md:70` proposes emitting `ESC[3J ESC[H ESC[2J` to subscribers on every resize, then says this is a no-op for plain shells. Plain shells are main-screen sessions with useful scrollback, so this would delete terminal history on every resize. The plan needs a narrower Claude/main-screen-TUI strategy, or a real terminal model/checkpoint approach, before implementation.

2. Medium: the sidecar launch path points at the wrong binary. `docs/impls/0011-pty-host-terminal-runtime.md:202` says Tauri invokes the sidecar via `std::env::current_exe()`, but `current_exe()` is the main app binary, not the `externalBin` sidecar. Unless the main binary intentionally multiplexes `--detach`, this launches the wrong executable. The plan should specify resolving the bundled sidecar path explicitly.

3. Low: the future-update list references a deleted file. `docs/impls/0011-pty-host-terminal-runtime.md:466` lists `docs/impls/0010-terminal-replay-cols-alignment.md`, but this PR deletes that file. Either keep `0010` as a superseded investigation note or remove it from the future update list.

Verification:
- `git diff --check origin/main...origin/pr/155` passed.

yicheng47:
Addressed in 674a5f4.

1. **High — resize-clear breaks shell scrollback.** You're right, the wire-level distinction doesn't exist. Dropped the live-path clear entirely. Live xterm now matches Terminal.app behaviour (claude-code's in-session redraws stack just like they do anywhere else — we deliberately don't fight that). The resize-stack fix moves into the snapshot path: host parses every PTY byte through `vt100::Parser`, and `Attach` returns `Screen::contents_formatted()` instead of the raw event tape. That makes the reattach view "exactly the visible region the agent is rendering right now" without touching the live stream. `vt100` is now v1 infrastructure, not a deferred optimization.

2. **Medium — `current_exe()` resolves the wrong binary.** Fixed. Sidecar is registered as `bundle.externalBin` and resolved via Tauri's `PathResolver` at runtime (`app.path().resolve("runner-pty-host", BaseDirectory::Resource)`). Dev and packaged-bundle paths called out explicitly.

3. **Low — orphaned 0010 reference.** Removed from Step 9's update list, with a note that 0010's diagnosis is preserved in the new "Resize-stack fix mechanism" section above.

yicheng47:
Second-pass review findings:

1. High: snapshot replay still lacks a geometry pre-size. `docs/impls/0011-pty-host-terminal-runtime.md:379` replays snapshot output before guaranteeing xterm is at `snapshot.cols` / `snapshot.rows`. The updated plan says `Attach` returns one synthetic formatted-output event, not resize events, so the frontend can still write a 200-col serialized grid into the default 80-col xterm. That reintroduces the replay geometry bug. Fix: resize xterm to `snapshot.cols` / `snapshot.rows` before writing any output, or include an initial resize event as the first snapshot event.

2. Medium: host resize flow omits resizing the headless terminal model. `docs/impls/0011-pty-host-terminal-runtime.md:296` resizes the PTY master and records the event, but does not resize the host-side `alacritty_terminal::Term`. The snapshot depends on the headless model matching the child PTY geometry; resizing only the PTY master leaves the parser grid stale after the first window resize. The resize flow should update `Term` before processing the child's SIGWINCH redraw.

3. Low: stale `vt100` API wording remains after switching to Alacritty. `docs/impls/0011-pty-host-terminal-runtime.md:396` still says `Screen::contents_formatted()` even though the plan switched from `vt100` to `alacritty_terminal`, and later correctly says Alacritty has no built-in ANSI serializer. Clean that up to avoid sending implementers toward a nonexistent API.

Prior findings look addressed.

Verification:
- `git diff --check origin/main...origin/pr/155` passed.

yicheng47:
Addressed in 82cdfcb.

1. **High — snapshot replay lacked a geometry pre-size.** Right call. The frontend replay algorithm in Step 5 now has an explicit step 4: resize xterm to `snapshot.cols` / `snapshot.rows` *before* writing any output. Uses the existing top-level `cols`/`rows` fields on `SessionTerminalSnapshot` — no protocol change needed. Cross-referenced from the Snapshot semantics section so the contract is clear from both sides.

2. **Medium — host resize omitted the headless terminal.** Fixed in Step 3. Resize order is now: `term.resize()` first (so the parser grid is ready), `master.resize()` second (so the child gets SIGWINCH into a parser that's already at the new geometry), then log + broadcast. Spelled out why the headless model must stay in lockstep with the PTY master — snapshot fidelity depends on it.

3. **Low — stale `Screen::contents_formatted()` wording.** Cleaned up the Snapshot semantics section to say "the host's `grid_to_ansi` serialization" instead. The one remaining mention in "Why alacritty_terminal for v1" is the deliberate compare-and-contrast ("alacritty does not expose a built-in serializer unlike vt100's `contents_formatted()`") — kept as-is since it's flagging the missing API as the cost of the swap, not directing implementers to call it.

yicheng47:
Third-pass review finding:

1. Medium: stale raw-tape replay language remains after the v1 design moved to headless-terminal snapshots. `docs/impls/0011-pty-host-terminal-runtime.md:131` still says the first implementation should replay the raw event tape on attach and defer checkpoint snapshots, but the updated v1 design says `Attach` returns `screen_to_ansi` output from the headless terminal, not raw tape. The same stale assumption shows up again in verification/non-goals around “all events through `last_seq`” and raw-tape replay. Clean those up so implementers do not rebuild the now-rejected raw replay path.

Prior second-pass findings look addressed: xterm pre-size is explicit, host `Terminal` resize is included, and the stale `Screen::contents_formatted()` directive is removed.

Verification:
- `git diff --check origin/main...origin/pr/155` passed.

yicheng47:
Addressed in 788c989.

**Medium — stale raw-tape replay language.** Three spots left over from when raw-tape replay was v1 and the headless emulator was deferred. All cleaned up:

1. **Approach section's "first implementation" paragraph** — now states the v1 design directly (headless terminal + `screen_to_ansi` snapshot) and explicitly rejects raw-tape replay with a one-liner on why ("faithful tape replay would re-stack every historical SIGWINCH redraw frame just like the failed PR #154 did with `tmux capture-pane`").
2. **Files-to-modify row for `RunnerTerminal.tsx`** — was "Replay output + resize events from the host tape"; now "Pre-size xterm to snapshot dims, then write the host's `screen_to_ansi` bytes; resume live events from `last_seq + 1`".
3. **Unit-test list and Non-Goals** — "Snapshot returns all events through `last_seq`" replaced with a proper snapshot bytes round-trip test (feed bytes into Terminal A → `screen_to_ansi` → feed back into Terminal B → assert grids and modes match). Non-goal "Do not optimize replay with terminal checkpoints until raw tape replay is correct" replaced with "Do not replay the raw event tape on `Attach` — the v1 path is `screen_to_ansi`. The tape exists for durability and debugging only".

Remaining "raw event tape" mentions in the doc are all in their durability/debugging role; framing is now consistent end-to-end.

yicheng47:
Fourth-pass review findings:

1. Medium: the plan still overclaims plain-shell snapshot fidelity. `docs/impls/0011-pty-host-terminal-runtime.md:102` says plain-shell snapshots are “lossless” because the parser tracks main-screen + scrollback, but Step 5 describes returning `screen_to_ansi` for the current state/visible screen. Recreating scrollback in xterm is not the same as painting the visible grid with cursor-addressed ANSI. The plan needs to choose: visible-region snapshot only, or full scrollback reconstruction with a concrete encoding strategy.

2. Low: stale verification wording still expects resize events in the snapshot. `docs/impls/0011-pty-host-terminal-runtime.md:636` says the replay snapshot should contain a resize event before later output. The v1 snapshot contract now says a single synthetic `Output` event plus top-level `cols`/`rows`. Update this verification item.

3. Low: protocol docs conflate live raw PTY bytes with synthetic snapshot bytes. `docs/impls/0011-pty-host-terminal-runtime.md:183` documents `Output.data` as raw PTY bytes, but snapshot output is synthetic `screen_to_ansi` bytes. Reword to “terminal byte stream” and distinguish live raw PTY output from synthetic snapshot output.

Verification:
- `git diff --check origin/main...origin/pr/155` passed.

yicheng47:
Branch now squashed to a single commit (e0a6342). Fourth-pass findings addressed inside that same commit:

1. **Medium — overclaimed plain-shell snapshot fidelity.** You're right, I conflated "parser tracks scrollback internally" with "snapshot includes scrollback". The "Resize-stack fix mechanism" section now says the v1 snapshot is **visible region only**, with explicit acknowledgement that plain shells lose pre-reload scrollback (e.g. an earlier `ls` that has scrolled out of view). Spells out why the naive fix doesn't work — walking the scrollback grid would re-stack every historical SIGWINCH redraw frame for main-screen TUIs like claude-code, and there's no wire-level signal distinguishing meaningful shell scrollback from stale TUI redraws. Tags this as the one v1 regression vs. tmux's `capture-pane -S - -E -`, with a Phase-2 follow-up noted (per-runtime flag or OSC annotation scheme) and a release-notes mention.

2. **Low — stale "resize event before later output" verification.** Replaced with a verification item that actually matches the v1 snapshot contract: "Resize session, then immediately `Attach`; verify `HostSnapshot.cols`/`rows` match the post-resize geometry and the serialized bytes in `events[0]` reflect the redrawn screen at the new dims" — with a parenthetical noting that v1 returns a single synthetic Output event plus top-level dims, no intra-snapshot Resize entry to assert on.

3. **Low — conflated live raw PTY bytes with synthetic snapshot bytes in protocol docs.** Rewrote the `TerminalReplayEvent::Output.data` comment to spell out both carriers explicitly: on `HostEvent::Output` (live path) it's raw PTY bytes forwarded from the child; on `HostSnapshot.events[0]` (attach path) it's synthetic bytes from `screen_to_ansi` over the headless `Terminal`'s current screen. Frontend write behaviour is the same in both cases ("write verbatim into xterm"), which is why one `data` field carries both, but the origin distinction is now clear.

PR #156, opened 2026-05-18
feat: in-process PTY terminal runtime (0011 umbrella)

yicheng47:
Umbrella PR for the 0011 migration. Sub-PRs target `feat/pty-host-runtime`; this PR merges to `main` when all 3 steps are in.

**The first two commits on this branch are the plan-doc evolution.** The design has gone through two pivots since the original plan landed on main:

- Original (merged as PR #155): headless emulator on host + screen_to_ansi snapshot
- Pivot 1 ([88f9382](../commit/88f9382)): drop the headless emulator, accept "no visible state on reattach"
- Pivot 2 ([e269891](../commit/e269891)): drop the sidecar entirely, run `PtyRuntime` in-process

See the latest [docs/impls/0011-pty-host-terminal-runtime.md](../blob/feat/pty-host-runtime/docs/impls/0011-pty-host-terminal-runtime.md) for the final design.

## Pivot summary

| Was (PR #157) | Was (rewrite 1) | Now |
|---|---|---|
| Host-side `alacritty_terminal::Term` | No host-side terminal model | No host-side terminal model |
| `screen_to_ansi` serializer on `Attach` | No snapshot | No snapshot |
| Detached sidecar (setsid + lockfile) | Managed-child sidecar | **No sidecar — in-process runtime** |
| IPC: Unix socket + framed JSON | IPC: Unix socket + framed JSON | **No IPC** — direct method calls |
| `runner-pty-host` external bin in bundle | `runner-pty-host` external bin | **No external bin** |
| Tries to restore visible state on reattach | Auto-attach + spawn-with-resume tree | **Manual Resume button only** |

## Step checklist

- [ ] **Step 1** — `PtyRuntime` (`src-tauri/src/session/pty_runtime.rs`) — `SessionRuntime` impl over `portable-pty`, in-process
- [ ] **Step 2** — Lifecycle wiring — startup DB cleanup + on-quit hook + Cmd+R re-attach + Resume button (existing `session_resume`)
- [ ] **Step 3** — Remove tmux — delete `TmuxRuntime`, `session/tmux.rs`

## What's dropped vs the original plan

- Sidecar binary (`runner-pty-host`) and `[[bin]]` registration
- Detach machinery (setsid, double-fork, lockfile, stale-socket probe)
- Protocol crate (`crates/runner-core/src/pty_host.rs`)
- IPC framing, `HostMessage` envelope
- `HostSnapshot`, `TerminalReplayEvent`, `screen_to_ansi`
[...]

PR #157, opened 2026-05-18
feat(pty-host): protocol crate + sidecar with host-owned sessions

yicheng47:
Sub-PR against the 0011 umbrella (#156). Covers **Steps 1, 2, and 3** of [docs/impls/0011-pty-host-terminal-runtime.md](../blob/feat/pty-host-protocol-sidecar/docs/impls/0011-pty-host-terminal-runtime.md).

### Step 1 — protocol crate (`crates/runner-core/src/pty_host.rs`)

| Type | Purpose |
|---|---|
| `HostRequest` | `Spawn` / `Attach` / `Input` / `Paste` / `Key` / `Resize` / `Stop` / `Status` / `List` |
| `HostResponse` | `Ack` / `Error` / `Spawned` / `SessionStatus` / `Snapshot` / `Sessions` |
| `HostEvent` | `Output` / `Resize` / `Exit` / `RunnerStatus` (push) |
| `HostMessage` | `{type: response \| event}` envelope so the Tauri side routes by top-level discriminator instead of inspecting inner `kind` tags |
| `TerminalReplayEvent` | `Output { seq, data }` / `Resize { seq, cols, rows }` |
| `HostSnapshot`, `SpawnSpecWire`, `HostSessionStatus` | structured payloads |

21 serde round-trip tests. All discriminators are `snake_case` as documented.

### Step 2 — sidecar daemon (`src-tauri/src/bin/runner-pty-host.rs`)

* `--detach`: `setsid` + double-fork, chdir(\"/\"), umask 0o077, redirect stdin/stdout to /dev/null (stderr stays open for dev runs)
* `--socket-dir <PATH>`: where the lockfile, socket, and per-session replay logs live (typically `<app_data>/pty-host`)
* `fs2::FileExt::try_lock_exclusive` on `pty-host.lock` — second invocation exits cleanly so the Tauri startup can fall back to connecting
* Unix socket listener at `<socket-dir>/runner-pty-host.sock` with 0o600 perms
* Stale-socket detection: probe-then-unlink only if no peer responds
* 4-byte big-endian length-prefixed JSON frames, capped at 16 MiB

### Step 3 — host-owned sessions (same file)

* `tty::new(&TtyOptions { shell, working_directory, drain_on_exit: true, env }, WindowSize, window_id)` spawns the PTY + child in one call (alacritty's own PTY layer; rustix_openpty under the hood)
* **Reader thread per session** reads raw PTY bytes from `pty.file().try_clone()`, feeds them through `Processor::advance(&mut term, &bytes)` against a shared `Arc<FairMutex<Term<HostListener>>>`, appends a `TerminalReplayEvent::Output` row to `terminal.ndjson` (durability tape — not on the replay path), and broadcasts the raw bytes to subscribers
* **`screen_to_ansi` serializer** with round-trip test: feed bytes into Term A → serialize → feed serialized into Term B → grids match cell-by-cell. Emits alt-screen toggle, hard reset, per-row cursor positioning, SGR runs (8 attrs + 16 named + RGB + indexed), restores cursor pos + visibility.
* `HostListener` — local `EventListener` impl that forwards `Event::PtyWrite` (DA1 / DSR responses) back into the child via the shared writer
* `resize` — orders matter: `Term::resize` first (so the parser grid is ready for the SIGWINCH redraw at new dims), then `pty.on_resize(WindowSize)`, then log + broadcast
* `stop` — `libc::kill(pid, SIGTERM)`; `ESRCH` (already dead) silently ignored
* Per-connection: separate reader + writer threads bridged by `mpsc<HostMessage>`. Subscribers from `Attach` get `tx.clone()`; failing sends self-cleanup via `subscribers.retain(...)` on next broadcast.

## Decisions made during implementation

These are reflected in the plan-doc diff in this PR:

1. **`wezterm-term` → `alacritty_terminal`**. `wezterm-term` is not published to crates.io (only a third-party `tattoy-wezterm-term v0.1.0-fork.5` is). A git dep on `wezterm/wezterm` would trade CI-clone cost + supply chain risk for a protocol-coverage edge we don't need yet.
2. **Drop `portable-pty`**. `alacritty_terminal::tty::new` already allocates the PTY + spawns the child. Zed uses exactly this combination at `zed/crates/terminal/src/terminal.rs:520`. A separate PTY crate would duplicate SIGWINCH / drain-on-exit / signal-handling surface.
3. **Skip alacritty's `EventLoop`**. Its reader feeds bytes straight into `Processor::advance` and never exposes them externally (only a compile-time `ref_test` file tap). The live xterm path needs raw bytes — using alacritty's loop would force us to either feature-fork it or run a parallel reader anyway, so we just write the reader.

## Test results

\`\`\`
cargo test --workspace
test result: ok. 42 passed; 0 failed; 0 ignored
\`\`\`

* runner-core: 21 protocol round-trip + tag-shape tests
* runner-pty-host: 7 tests covering frame I/O, base64 known vectors, ULID shape, screen_to_ansi round-trip, dispatch::List against empty registry
* runner_lib + event_log: pre-existing tests still pass

## What's NOT in this PR

* No Tauri wiring — `PtyHostRuntime` lands in Step 4 (next sub-PR)
* No frontend changes — Step 5 territory
* No `bundle.externalBin` registration / `stage-runner-pty-host.mjs` — Steps 6+7 ship the bundling
* Paste verification still routes through the existing tmux path — Step 8

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,6 +41,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "alacritty_terminal"
+version = "0.26.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bda177466b9524d59f1b12f0dd30b68696788e9992a7e959021c4a0ed96fcf59"
+dependencies = [
+ "base64 0.22.1",
+ "bitflags 2.11.1",
+ "home",
+ "libc",
+ "log",
+ "miow",
+ "parking_lot",
+ "piper",
+ "polling",
+ "regex-automata",
+ "rustix",
+ "rustix-openpty",
+ "serde",
+ "signal-hook 0.4.4",
+ "unicode-width",
+ "vte",
+ "windows-sys 0.59.0",
+]
+
+[[package]]
 name = "alloc-no-stdlib"
 version = "2.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -872,6 +897,12 @@ dependencies = [
  "quote",
  "syn 2.0.117",
 ]
+
+[[package]]
+name = "cursor-icon"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f27ae1dd37df86211c42e150270f82743308803d90a6f6e6651cd730d5e1732f"
 
 [[package]]
 name = "darling"
@@ -1831,6 +1862,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
 
 [[package]]
+name = "home"
+version = "0.5.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc627f471c528ff0c4a49e1d5e60450c8f6461dd6d10ba9dcd3a61d3dff7728d"
+dependencies = [
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "html5ever"
 version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2545,6 +2585,15 @@ checksum = "50b7e5b27aa02a74bac8c3f23f448f8d87ff11f92d3aac1a6ed369ee08cc56c1"
 dependencies = [
  "libc",
  "wasi 0.11.1+wasi-snapshot-preview1",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "miow"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "536bfad37a309d62069485248eeaba1e8d9853aaf951caaeaed0585a95346f08"
+dependencies = [
  "windows-sys 0.61.2",
 ]
 
@@ -3720,6 +3769,7 @@ dependencies = [
 name = "runner"
 version = "0.1.14"
 dependencies = [
+ "alacritty_terminal",
  "base64 0.22.1",
  "chrono",
  "fs2",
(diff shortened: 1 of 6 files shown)


yicheng47:
Review findings:

1. High: attach sequence handling can drop live bytes after replay. In `src-tauri/src/bin/runner-pty-host.rs:537`, `seq.fetch_add(1)` returns the emitted event seq, so the atomic stores the next seq. But `attach_session` returns `last_seq = handle.seq.load()` at `src-tauri/src/bin/runner-pty-host.rs:597`. After one output event `seq=0`, attach returns `last_seq=1`; the next live output also gets `seq=1` and the frontend merge rule (`seq > last_seq`) will drop it. Also, subscription is pushed after snapshot capture, so bytes between snapshot and subscriber registration can be missed. Fix by making the atomic semantics explicit (`next_seq` vs `last_emitted_seq`) and making attach subscribe/buffer atomically with snapshot capture.

2. High: `Key` and `Paste` do not implement the protocol contract. The protocol says paste uses bracketed-paste semantics and key is a symbolic key translated by the host, but `dispatch` collapses paste to raw input at `src-tauri/src/bin/runner-pty-host.rs:356`, and `Key` writes `key.as_bytes()` at `src-tauri/src/bin/runner-pty-host.rs:382`. When Step 4 maps `SessionRuntime::send_key("Enter")` to this, it will type `enter` into the agent instead of sending CR. Multi-line paste can also submit line-by-line instead of a single bracketed paste.

3. Medium: snapshot replay restores the visible grid but not terminal modes needed for post-attach interaction. `screen_to_ansi::serialize` at `src-tauri/src/bin/runner-pty-host.rs:939` restores alt-screen and cursor visibility only. It does not emit APP_CURSOR, BRACKETED_PASTE, mouse reporting, focus reporting, application keypad, etc. After reattach, xterm may display the right frame but encode arrows/paste/mouse differently until the agent redraws those modes.

Verification:
- `git diff --check origin/feat/pty-host-runtime...origin/pr/157` passed.
- `cargo test -p runner-core` passed: 42 tests.
- `TAURI_CONFIG='{{"bundle":{{"externalBin":[]}}}}' cargo test -p runner --bin runner-pty-host` passed: 7 tests.
- Plain `cargo test --workspace` failed in a clean worktree because Tauri could not find the staged `binaries/runner-cli-aarch64-apple-darwin` external binary.

yicheng47:
Superseded by pivot — the headless-emulator path (alacritty Term + screen_to_ansi + mode restoration + key translation + paste bracketing) was hitting one subtle correctness finding per review pass and we kept adding layers to fix each one. Pivoting back to portable-pty + raw byte stream; xterm.js owns the terminal model. Trade-off: no visible state on app-restart reattach (agents still survive, view starts fresh). Plan rewrite incoming on the umbrella PR #156.

PR #175, opened 2026-05-20
fix(session): claude-code launch gate + non-blocking mission spawn

### LINKED ISSUE #171: bug: concurrent claude OAuth refresh race forces relogin on mission spawn / Resume All
## Description

When a mission spawns multiple `claude-code` runners in parallel (e.g., a 3-slot crew like architect/impl/reviewer, or a 7-slot werewolf game), Claude Code sometimes prompts for relogin in one or more of the panes. Same symptom on **"Resume all"** in `MissionWorkspace` and on resuming several direct chats in quick succession. Outside Runner — running one `claude` at a time in Terminal — relogin is never required.

Working hypothesis: a concurrent OAuth refresh-token race across the N parallel `claude` processes.

- Claude Code stores its credential in the macOS Keychain (`security find-generic-password -s "Claude Code-credentials"` confirms a `genp` item; no `~/.claude/.credentials*` file exists).
- Runner itself never touches the Keychain (verified: no `keychain`/`security`/`SecKey`/`keytar` references in `src-tauri/src/` or `src/`; no entitlements in `tauri.conf.json`). The Keychain ACL evaluates the calling process — `claude` (Anthropic-signed, Team `Q6L2SF6YDW`) — not Runner.app, so the parent app's identity is irrelevant for ACL.
- When N claudes start at once, each reads the same refresh token, and if it's close to expiry each POSTs to the OAuth refresh endpoint with that token. Refresh tokens are conventionally single-use: the first POST wins and gets a rotated token; the rest get `invalid_grant` and fall back to interactive login.
- Cross-check from prod logs: `~/Library/Logs/com.wycstudios.runner/runner.log` shows `session spawn:` lines for three runners landing in the same second on a mission start, and the keychain item's `mdat` (modification time) on a known relogin matched a mission spawn timestamp to the second — i.e., claude *did* rewrite the credential post-relogin from a Runner-spawned PTY.
- The fact that it happens "sometimes" fits: the race only matters when the cached token is close enough to expiry that all parallel claudes decide to refresh on this launch.

## Expected behavior

Spawning N `claude-code` PTYs in parallel (mission start, Resume All, or rapid per-slot resume) should not force the user to re-authenticate. The refresh, when needed, should happen once across all spawning claudes.

## Steps to reproduce

Reliable repro depends on token-expiry timing, but the following maximizes the odds:

1. Don't open `claude` for ~24h (let the cached token age toward expiry).
2. Start a multi-runner mission (e.g., the default build crew: architect + impl + reviewer) — or click **Resume all** on a mission whose slots are stopped.
3. Observe: one or more panes prompts for relogin in the TUI; the others succeed.

Negative control: spawning a single direct chat (one claude) under the same conditions does **not** trigger relogin.

## Relevant code

- `src-tauri/src/session/manager.rs:578` `SessionManager::spawn` — invoked once per slot for mission start, mission reset, and direct-chat spawn.
- `src-tauri/src/session/manager.rs:999` `SessionManager::resume` — invoked once per slot for per-slot resume, "Resume all", and rapid direct-chat resume.
- `src/pages/MissionWorkspace.tsx:338` `resumeMission` — frontend "Resume all" loop. `await` in the `for` only serializes PTY *creation*; once each PTY is alive, the claude inside it does its own OAuth refresh on its own timeline, so the actual refreshes still overlap. A frontend-only stagger does not fix this and only covers one of the affected entry points.
- `src-tauri/src/session/pty_runtime.rs:86` `PtyRuntime::spawn` — returns as soon as `spawn_command` starts the child; the suspected token refresh happens *inside* the child after this returns, so a mutex around `runtime.spawn()` alone is insufficient.
- `src-tauri/src/shell_path.rs` — the existing `HTTPS_PROXY`/`HTTP_PROXY`/`ALL_PROXY`/`NO_PROXY` capture (#160) is unrelated to this bug; that fix addresses proxy-via-rc-file users, not the OAuth refresh race.

## Proposed mitigation

Centralized **claude-code launch gate** in `SessionManager`, sitting just above the runtime spawn call:

- Scoped to `runner.runtime == "claude-code"` only. Other runtimes (codex, etc.) bypass the gate.
- Applied uniformly across every spawn + resume entry point listed above so all five entry points (mission start, mission reset, per-slot resume, "Resume all", direct-chat spawn/resume) share the same serialization.
- Serializes claude launches with a configurable conservative spacing — the spacing has to outlast not just `runtime.spawn()` but also the post-spawn window during which the child performs its OAuth refresh. Either hold the lock for a small grace period after `spawn()` returns, or wait for a stronger readiness signal from claude's stdout before releasing the next launch.
- Logs when the gate applies so the impact is visible in `runner.log`.
- Removable / relaxable once Anthropic addresses concurrent refresh upstream.

Rejected alternatives:
- Frontend stagger in `MissionWorkspace.resumeMission` — only covers "Resume all"; misses the other four entry points.
- Mutex on `runtime.spawn()` alone — releases before the child does its refresh; doesn't actually serialize the racing operation.
- Prewarm via `claude --version` at app start — almost certainly fully local, no network call, so would not refresh the credential. No other claude subcommand is obviously safe to fire as a prewarm probe without touching the user's actual session state.
- File upstream with Anthropic — orthogonal to the local mitigation; worth doing in parallel but doesn't ship a fix on our side.

## Environment

- OS: macOS (darwin 24.6.0)
- Version: 0.1.18
- Runtime: in-process portable-pty (impl 0011)

yicheng47:
## Summary

End-to-end fix for the concurrent claude-code OAuth refresh race surfaced in #171.

- **Launch gate** in `SessionManager` — deadline-based serializer for `claude-code` PTY spawns. First spawn pays 0; concurrent spawns 1.5s apart. Codex / shell / resume bypass.
- **Lead-first spawn order** so the lead always wins the no-wait pass.
- **Pill clears on TUI-ready** (`\x1b[?2004h` bracketed-paste enable, captured empirically from claude-code + codex startup) instead of first-reply-done.
- **`mission_start` / `mission_reset` no longer block on the gate** — register loop synchronous, gate-blocked PTY spawn dispatched as a background task. Start-mission RPC returns in ~50ms; pills cover the staggered boot.

## Test plan

- [x] `cargo fmt --all --check` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo test --workspace --all-targets` — 239 backend + crate tests passing (+9 new: gate math, gate accept/bypass, lead-first ordering)
- [x] `pnpm tsc --noEmit` clean
- [x] `pnpm eslint 'src/**/*.{ts,tsx}'` clean (no new warnings)
- [ ] Manual: start Build squad (3 claudes) — modal closes instantly; lead pane clears pill ~500ms after mount; workers ~1.5s/3s later; gate log lines visible in `runner.log`
- [ ] Manual: Resume-all over 3 stopped claude slots — no relogin
- [ ] Manual: Codex pane in a mixed-runtime mission — no `claude-code launch gate` log lines

Closes #171.

### DIFF
--- a/docs/features/21-resume-cwd-sessions.md
+++ b/docs/features/21-resume-cwd-sessions.md
@@ -1,0 +1,85 @@
+# 21 — Detect and resume existing agent sessions for the cwd
+
+> Tracking issue: [#176](https://github.com/yicheng47/runner/issues/176)
+
+## Motivation
+
+A lot of Runner's target users already use `claude-code` and `codex` directly in their terminal *before* they install Runner. When they then start a direct chat in Runner against the same working directory, they get a fresh agent with none of the conversation history they just built up. Today the workaround is to keep using the agent CLI outside Runner — exactly the split-attention problem Runner is trying to solve.
+
+Both agent runtimes already store per-project session history on disk and expose a resume verb (`claude --resume <uuid>` / `codex resume <uuid>`). Runner just needs to surface those sessions in the Start Chat modal and pass the right flag at spawn.
+
+## Scope
+
+### In scope (v1)
+
+- **Detection.** When the user picks a runner + cwd in the Start Chat modal, scan the runtime-specific on-disk session store for sessions whose recorded cwd matches. Surface them in a "Recent sessions for this directory" panel below the rest of the modal fields.
+- **Picker UX.** Each session row shows: relative timestamp ("3h ago", "yesterday"), one-line preview from the first user message, and the session id (dimmed). Sorted newest first; cap at the 10 most recent. Default selection is "Start fresh"; clicking a row pre-selects it for resume.
+- **Resume.** On submit, if a session was selected, pass the runtime's resume flag to the spawned child:
+  - claude-code: `claude --resume <uuid>` (the bundle's existing `agent_session_key` capture flow already proves this works in Runner; we just hand it the uuid instead of letting the agent pick).
+  - codex: `codex resume <uuid>` (codex uses positional uuid; `--last` is also supported but the picker makes the explicit form natural).
+- **Runtime coverage.** Both `claude-code` and `codex` in v1. `shell` runner type has no session concept and skips the panel entirely.
+
+### Out of scope (v1)
+
+- **Mission spawn.** Detection only surfaces in Start Chat. Missions stay fresh-start — multi-slot resume is a different shape (per-slot picker × N) that doesn't fit the modal cleanly. Revisit when there's user demand.
+- **Importing the transcript into Runner's event feed.** The resumed agent renders its own history into the PTY (xterm.js paints it like any other output); we don't replay it into Runner's NDJSON log. This means Runner's scrollback ring only sees post-resume bytes, but the user sees the full transcript in the terminal as the agent restores it.
+- **Cross-runtime resume.** No "resume a codex session in claude-code" — they're not on-the-wire compatible.
+- **Edit / delete from Runner.** Sessions remain owned by the agent CLI. If the user wants to delete a session, they do it from the agent's own surface.
+
+## Implementation phases
+
+### Phase 1 — disk layout adapters
+
+Add `session_store::SessionEntry { id, cwd, started_at, first_user_msg }` and a trait / function pair per runtime that lists entries for a given cwd:
+
+- **claude-code:** sessions live at `~/.claude/projects/<encoded-cwd>/<uuid>.jsonl`. The encoded-cwd convention is the absolute path with `/` replaced by `-` (e.g. `/Users/jason/foo` → `-Users-jason-foo`). Each JSONL's first non-system entry is the first user message — read just the head of the file (cap at a few KB).
+- **codex:** sessions live at `~/.codex/sessions/*.jsonl` (flat, not partitioned by cwd). Each session's JSONL records the cwd in its first event; we have to read all session heads and filter. Cache the index in memory across modal opens so we don't re-stat the whole dir on each render.
+
+Both adapters return `Vec<SessionEntry>` sorted by `started_at` desc, capped at N=10.
+
+### Phase 2 — `session_list_resumable` Tauri command
+
+```ts
+session_list_resumable({ runner_id, cwd }) → Promise<SessionEntry[]>
+```
+
+Looks up the runner template's runtime, dispatches to the right adapter, returns the entries. Errors (missing home dir, unreadable session file) degrade silently to an empty list — never surface a "couldn't list sessions" toast; the user just sees "no recent sessions."
+
+### Phase 3 — frontend: Start Chat modal panel
+
+- Below the existing fields (runner picker, working dir, title), add a `<RecentSessionsPanel>` that calls `session_list_resumable` whenever runner + cwd both have values.
+- Default state: "Start fresh" radio is selected. Each session row is a radio option below it. Loading state shows a small spinner; empty state shows "No prior sessions for this directory."
+- On submit: if a session radio is selected, pass `resume_session_id: <id>` to the existing `session_spawn_direct` command. Otherwise the field is absent and spawn behaves exactly as today.
+
+### Phase 4 — backend: thread the resume id through spawn
+
+- `session_spawn_direct` grows an optional `resume_session_id: Option<String>`.
+- The runtime adapter (`router::runtime::*`) extends its arg-building to prepend the resume flag when set. claude-code: `--resume <uuid>`. codex: positional `resume <uuid>`.
+- Persist the chosen uuid on the new `sessions` row as `agent_session_key` (the column already exists for codex auto-capture; reuse it). This lets the existing cross-restart reattach path resume the same agent session if the user restarts the app mid-chat.
+
+### Phase 5 — empty / edge cases
+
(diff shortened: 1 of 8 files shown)


yicheng47:
Review findings:

1. High: pending background spawns survive Stop/Archive/Reset. `mission_start` queues `complete_mission_session_spawn` in a detached `spawn_blocking` task, but pending sessions are not in `SessionManager.sessions` until the PTY is actually spawned. `mission_stop` / `mission_archive` only call `kill_all_for_mission`, which only sees live handles. A user can start a gated multi-Claude mission, immediately stop/archive/reset, and the remaining pending slots will still spawn afterward into a stopped/archived/reset mission. Refs: `src-tauri/src/commands/mission.rs:792`, `src-tauri/src/session/manager.rs:873`, `src-tauri/src/session/manager.rs:2014`.

2. Medium: async spawn failure updates DB but does not notify the mounted workspace. The background task marks a failed pending spawn `crashed`, but emits no `session/exit` or other event. `MissionWorkspace` only refreshes session rows on `session/exit`, so the current UI can keep showing the slot as running/starting and never expose the Resume button until a manual refresh. Refs: `src-tauri/src/commands/mission.rs:802`, `src/pages/MissionWorkspace.tsx:209`.

3. Medium: direct-chat resume can clear the loader from stale pre-resume output. `resumeChat` sets `resuming=true` before calling `api.session.resume`; the new snapshot fast-path can read the old stopped session buffer before the backend purges it, see an old `\x1b[?2004h`, and clear the resuming overlay before the new PTY exists. Refs: `src/pages/RunnerChat.tsx:346`, `src/pages/RunnerChat.tsx:623`, `src-tauri/src/session/manager.rs:1528`.

Validation note: I no longer consider ungated `claude-code` Resume All a bug if `claude --resume <id>` only loads local conversation state and does not touch Anthropic/auth until the next user turn. Since Runner does not auto-send a first turn on resume, bypassing the launch gate on `SessionManager::resume` is fine under that behavior. It would be worth keeping the manual Resume All check in the PR test plan because issue #171 explicitly named that path.

yicheng47:
Review findings:

1. High: pending spawn cancellation can be lost across reset/restart batches. In `src-tauri/src/session/manager.rs:2135`, `drop_pending_mission_cancel` removes the cancellation flag by `mission_id` only. If an old background spawn task is still draining, then `mission_reset` registers a new flag for the new spawn batch, the old task can later call `drop_pending_mission_cancel(&mission_id)` and remove the new batch flag. A subsequent Stop/Archive/Reset then cannot cancel the new pending queue, so slots can still spawn after the mission was stopped. Make the drop conditional on the exact `Arc` owned by the task, or use a per-batch generation/token.

2. High: Stop/Archive/Reset still has a spawn race after `runtime.spawn`. `complete_mission_session_spawn` checks `cancel` before the gate and after the gate (`src-tauri/src/session/manager.rs:887` and `src-tauri/src/session/manager.rs:907`), but there is no check after `runtime.spawn(spec)` returns at `src-tauri/src/session/manager.rs:917` and before inserting the handle at `src-tauri/src/session/manager.rs:946`. If Stop fires while `runtime.spawn` is in progress, `kill_all_for_mission` cannot see the session yet; once spawn returns, the PTY is inserted and forwarded anyway. Check cancellation immediately after spawn returns and tear down the runtime session before inserting, or otherwise make in-flight spawns visible to `kill_all_for_mission`.

3. Medium: direct-chat resume can still clear from a stale snapshot. `src/pages/RunnerChat.tsx:623` sets `resuming` before awaiting `api.session.resume`, while the resuming effect independently calls `api.session.outputSnapshot` at `src/pages/RunnerChat.tsx:383`. Tauri commands can run concurrently, so that snapshot can still beat the backend purge at `src-tauri/src/session/manager.rs:1405` and see the old TUI-ready bytes. Gate the snapshot fast path until `session_resume` resolves, skip the snapshot fast path during resume, or add a resume generation/epoch.

Note: I am no longer treating Resume All as a blocker here, assuming resume does not trigger backend interaction.

Verification:
- `cargo test --workspace --all-targets` passed on PR head `070d5f8`.
- `pnpm run lint` passed with Fast Refresh warnings in `SessionEndedOverlay.tsx` and `UpdateContext.tsx`.

yicheng47:
Review finding:

1. High: `runner_delete` still misses gated fresh spawns. `src-tauri/src/commands/runner.rs:655` only calls `kill_all_for_runner`, and `src-tauri/src/session/manager.rs:2211` only scans live `SessionHandle`s. But this PR now inserts fresh session rows before the Claude launch gate, while direct/mission PTYs may not be in `SessionManager.sessions` yet. Deleting a runner during that window cascades the DB row via `sessions.runner_id ON DELETE CASCADE`, but the in-flight spawn keeps its cloned `Runner`, ignores the zero-row runtime metadata update at `src-tauri/src/session/manager.rs:948`, then inserts a live handle at `src-tauri/src/session/manager.rs:970`. Same shape exists in direct chat `spawn_direct`. Result: a PTY can start for a deleted runner with no DB row. Check the session row still exists after the gate and again after `runtime.spawn`; if missing, stop the just-spawned runtime session and do not insert the handle.

Validation:
- `cargo test --workspace --all-targets` passed on PR head `9dcb696`.
- `pnpm run lint` passed with existing Fast Refresh warnings.
- `git diff --check origin/main...origin/pr-175-review` passed.

PR #191, opened 2026-05-22
fix(terminal): force agent redraw on tab activation (#177) + editable crew goal

### LINKED ISSUE #177: bug: terminal pane renders black with scattered text after switching back to a tab that accumulated output while hidden
## Description

After switching away from a runner's PTY tab and letting the agent
produce a non-trivial amount of output while that tab is hidden,
switching back to the tab lands on a mostly-black canvas with scattered
fragments of text at arbitrary positions — not the agent's current
rendered state. The pane stays in that broken state until any resize
event nudges xterm to refit, at which point the agent's SIGWINCH-driven
redraw lands and the canvas repaints correctly.

Originally filed as codex-specific, but the symptom is not bound to any
one runtime — it reproduces for every TUI-style runtime (`codex`,
`claude-code`, `gemini`, etc.) once enough output has accumulated while
the pane was hidden. `claude-code` masks the symptom slightly (alt-
screen reattach handling + the bracketed-paste TUI-ready signal cover
the cleanest case, where the agent is parked at a prompt with no
overflow), so the report initially landed as codex-only.

The defining triggers are:

1. The session has produced more output than fits in the current
   viewport (i.e., enough to overflow / scroll while hidden).
2. The pane was not the active tab while that output was emitted.
3. The user then switches back to the pane.

## Expected behavior

Switching to an already-open PTY tab should drop straight into the
live xterm with its current content painted correctly — the same as
returning via window-focus or document-visibility, both of which
already work.

## Steps to reproduce

1. Start a mission with at least one TUI runner slot (any of
   `codex`, `claude-code`, or `gemini` reproduce).
2. Get that runner busy enough that it's producing PTY output that
   exceeds the visible viewport (long agent turn, file dumps, build
   logs, anything that scrolls).
3. While that output is streaming, click another tab (Feed, or any
   other PTY) to navigate away — the source pane goes `display:none`
   via `MissionWorkspace.tsx`'s `Pane` wrapper.
4. Wait a few seconds; let more output accumulate.
5. Click back to the original pane.

Observed: black canvas with scattered, mispositioned text fragments
instead of the agent's current rendered state.

Workaround: drag the window edge a pixel, toggle a side panel — any
resize event causes `fit.fit()` + a SIGWINCH to the agent, and the
agent's next full redraw paints correctly.

## What's happening

This is *not* the same as the original framing (\"the activation
effect doesn't call `refreshTerm()`\"). The activation effect at
`src/components/RunnerTerminal.tsx:656–707` already calls
`fit.fit()` + `webglRef.clearTextureAtlas()` + `t.refresh(0, t.rows - 1)`
+ `t.focus()` on the false→true `active` transition, and (when dims
diverge) re-pushes geometry to the backend, which sends SIGWINCH.
That landed in #163 / #167. The bug survived because the deeper
issue is in what the buffer contains by the time the activation
effect fires:

- Once snapshot replay is done (`replayDoneRef.current = true`), live
  PTY output writes straight into the xterm buffer regardless of
  whether the pane is mounted-and-active or mounted-and-hidden
  (`RunnerTerminal.tsx:605–607`).
- While the pane is `display:none`, the container's rect is 0×0;
  `refitAndPush` bails on the rect guard (`:430–432`). Meanwhile the
  WebGL renderer skips frame rendering because the canvas has no
  layout — but xterm keeps ingesting bytes into its internal
  buffer at the cols/rows it last had.
- If the window or a side panel was resized while the pane was
  hidden, the cols/rows xterm should have on return differ from the
  ones the buffer was filled at. When `fit.fit()` resizes on
  activation, the already-written cells reflow against new dimensions
  and absolute-positioned glyphs land at wrong positions — the
  \"scattered text\" effect.
- Even with no intervening resize, the buffer can be mid-frame: the
  agent was halfway through emitting a redraw sequence (clear,
  move-to, paint…) when the user switched back, and `refresh()` just
  redraws whatever partial state landed.
- `refresh()` repaints the *current viewport* from the buffer. It
  doesn't ask the agent to redraw. The agent only redraws on its own
  schedule or in response to a SIGWINCH it actually receives —
  pushSize is deduped against the previously-pushed cols/rows
  (`:694–697`), so a same-dims activation doesn't kick the agent.

The \"resize fixes it\" workaround works because the manual resize
forces fit to produce different dims, the dedup guard releases, and
the backend SIGWINCH lands → agent redraws fully.

## Possible fix surfaces

- **Don't ingest live bytes while the pane is hidden.** Park them
  on `pendingLiveRef` (the path already used for pre-activation
  replay) until the pane comes back, then drain into a freshly-fit
  terminal. Most surgical; preserves scrollback semantics. Cost:
  buffering bytes for potentially-long hidden intervals.
- **Force an unconditional SIGWINCH on activation**, bypassing the
  cols/rows dedup. Cheap; agent always redraws fully when the user
  returns. Cost: one redraw per tab switch even when nothing changed.
- **Reset xterm + drain snapshot on every activation**, treating
  every return as a fresh attach. Heaviest; same machinery as initial
  mount. Likely overkill.

The middle option is probably the right first step.

## Relevant code

- `src/components/RunnerTerminal.tsx:605–607` — live output writes to
  the xterm buffer with no active-tab gate; this is the path that
  fills hidden panes with mid-stream bytes.
- `src/components/RunnerTerminal.tsx:656–707` — activation effect
  (fit + atlas clear + refresh + focus + conditional pushSize). The
  conditional pushSize dedup at `:694–697` is what suppresses the
  fix-by-SIGWINCH path on same-dim returns.
- `src/components/RunnerTerminal.tsx:425–449` — `refitAndPush` +
  ResizeObserver path; bails on 0×0 rect while pane is hidden, so
  off-screen layout changes don't propagate to xterm dims until
  activation.
- `src/pages/MissionWorkspace.tsx:1150–1172` — `Pane` wrapper hides
  inactive tabs via Tailwind `hidden` (`display:none`), which is the
  zero-rect state that triggers the bail above.

## Environment

- OS: macOS (darwin 24.6.0)
- Version: 0.1.20
- Reproduces across all TUI runtimes once enough output accumulates
  while the tab is hidden. Previously framed as codex-only because
  claude-code's alt-screen + TUI-ready handling masks the cleanest
  case.

yicheng47:
## Summary

- **Fixes #177** — switching back to a runner's PTY tab after it had accumulated output while hidden showed a mostly-black canvas with scattered text. The activation effect now does an unconditional viewport wipe (TUI runtimes) + `(cols, rows-1) → (cols, rows)` SIGWINCH dance, replacing the dim-deduped pushSize. Kernel TIOCSWINSZ dedup means a same-size resize is a no-op, so the dance is required to force the agent to repaint. Rows-perturbation (not cols) — cols would deposit `claude-code`'s explicit-`\n` narrow lines into scrollback.
- **Resume contract fix** — `RunnerTerminal.measure()` returns last-fit dims on `0×0` rect when non-default, and `MissionWorkspace.resumeMission` pre-walks for any measurable terminal's dims and uses them as `sharedDims` fallback. Stops the resume RPC from sending `(null, null)` for hidden tabs (which made the backend spawn at 80×24 and the agent paint its `--resume` conversation history narrow).
- **Editable crew Default goal** — `CrewEditor` now has a section between Purpose and Team conventions for editing the crew's default goal, mirroring the addendum section pattern. Backend already supported it; the affordance just wasn't wired up.

Impl plan: `docs/impls/0012-force-redraw-on-tab-activation.md`.

## Test plan

- [ ] Start a mission with a `codex` and a `claude-code` runner. Get one busy producing output. Switch to another tab. Wait ~5s. Switch back → expect a brief black flash then a clean repaint, not the scattered-text state.
- [ ] Plain shell runner: `for i in {1..200}; do echo "line $i"; sleep 0.05; done`. Switch away while running, switch back → scrollback intact, no flash.
- [ ] Resize the app window between switch-away and switch-back. Resume should land at new dims with no residual content from the old dims.
- [ ] Toggle the right rail while the source pane is hidden, switch back → pane refits and repaints correctly.
- [ ] Rapid tab switching (A → B → A → C → A) → each return kicks a fresh redraw, no race-induced stale state.
- [ ] Stop a mission, then click Resume from the **feed tab** (not a slot tab). Switch to a paused slot tab and scroll up → conversation history should be at the full container width, not 80 cols.
- [ ] Dead session — kill a runner from its tab, switch away, switch back to the dead tab → no console errors, no thrown promise.
- [ ] Open a crew that has a Default goal set on creation — verify the new section displays it and has an Edit button.
- [ ] Open a crew with no goal — verify the "+ Add default goal" placeholder + hint.
- [ ] Edit the goal, Save → reload the crew page, confirm persistence. Clear the goal (empty text) + Save → confirm it reverts to the placeholder state.

### DIFF
--- a/docs/impls/0012-force-redraw-on-tab-activation.md
+++ b/docs/impls/0012-force-redraw-on-tab-activation.md
@@ -1,0 +1,205 @@
+# Force Agent Redraw on Tab Activation
+
+> Fixes [#177](https://github.com/yicheng47/runner/issues/177).
+
+## Context
+
+Switching back to a runner's PTY tab after it has accumulated output
+while hidden shows a mostly-black canvas with scattered text fragments
+instead of the agent's current rendered state. A manual window resize
+fixes it. Reproduces for every TUI runtime (`codex`, `claude-code`,
+`gemini`, …) once enough output landed while the tab was inactive.
+
+The activation effect at `src/components/RunnerTerminal.tsx:656–707`
+already does the right local-side work — `fit.fit()`, atlas clear,
+`term.refresh()`, focus — but `refresh()` only repaints the current
+xterm buffer. The buffer itself is the problem: while the pane was
+hidden, live PTY bytes wrote straight into it
+(`RunnerTerminal.tsx:605–607`). By the time the user returns the
+buffer is either mid-frame (the agent was halfway through a redraw at
+the moment of switch-away) or dim-mismatched (the window or a side
+panel resized while the pane was hidden, so xterm's grid no longer
+matches what the buffer was painted at). Refreshing it just shows the
+broken state.
+
+The pushSize path inside the activation effect is deduped against the
+previously-pushed cols/rows (`:694–697`) — a same-dim return never
+re-pushes, so the agent never receives a SIGWINCH and never knows to
+repaint.
+
+The "drag the window edge a pixel" workaround works because a real
+dim change releases the dedup, the backend resize fires, and the
+agent's full SIGWINCH redraw lands.
+
+## Approach
+
+Force the agent to repaint on every tab activation by doing a
+SIGWINCH dance (resize one row below current → resize back) so the
+kernel's TIOCSWINSZ dedup can't suppress the signal. Pair it with a
+viewport wipe for TUI runtimes so the user sees a clean black canvas
+during the few-ms gap before the agent's redraw lands, instead of the
+mid-frame mess.
+
+Why a dance and not just an unconditional `api.session.resize`: both
+Linux and macOS pty drivers compare the incoming `struct winsize`
+against the cached value and only signal `SIGWINCH` when the bytes
+differ
+(`drivers/tty/tty_io.c:tty_do_resize` on Linux,
+`bsd/kern/tty.c:ttioctl_locked` TIOCSWINSZ branch on Darwin). A
+single-call same-size resize is a no-op at the kernel level. Two
+resizes — `(cols, rows-1)` then `(cols, rows)` — guarantee the cached
+winsize changes between calls, so each ioctl produces a SIGWINCH.
+
+Why we perturb **rows**, not **cols**. An earlier draft of this plan
+used `(cols-1, rows) → (cols, rows)`. That worked for the alt-screen
+case (claude-code's alt buffer) but corrupted scrollback under any
+main-screen TUI mode: claude-code (and similar agents) wrap their
+own text by emitting explicit `\n` at their computed cols boundary,
+not relying on terminal auto-wrap. The intermediate `(cols-1)` paint
+deposited hard-wrapped narrow lines into xterm's buffer. xterm can
+only soft-reflow width-wrapped lines; explicit newlines stick. Every
+tab return then layered another stripe of narrower lines into
+scrollback, visible on scroll-up even though the current viewport
+looked fine. Row perturbation keeps content width at `cols`
+throughout — the `(rows-1)` intermediate paint is just one line
+shorter, and the second SIGWINCH's `\x1b[2J`-led repaint at the
+final row count cleans it up. No width pollution.
+
+Tradeoff: a one-cycle `(rows-1)`-tall intermediate frame may flash
+through the buffer before the second SIGWINCH's `rows`-tall repaint
+lands. The viewport wipe hides this for TUI runtimes. For plain
+shells (no `runtime` in `runtimeClearsOnResize`), the intermediate
+is harmless — shells don't redraw on SIGWINCH.
+
+---
+
+## Step 1: Replace the deduped pushSize in the activation effect with a SIGWINCH-guaranteeing dance
+
+**File: `src/components/RunnerTerminal.tsx`**
+
+In the activation effect (`useEffect` at `:650`), replace the
+conditional pushSize block at `:684–703` with an unconditional
+viewport wipe + dance. The wipe matches the existing pushSize behavior
+at `:417–421`: clear the visible region only for runtimes where
+`runtimeClearsOnResize` is true (claude-code, codex; plain shells
+keep their history).
+
+Replace the conditional pushSize block with an unconditional wipe +
+rows-perturbing dance:
+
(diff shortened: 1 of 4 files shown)


yicheng47:
Review findings:

1. Medium: feed-tab mission resume can still spawn resumed PTYs at 80x24. `src/pages/MissionWorkspace.tsx:397` looks for a shared fallback by calling `measure()` on the mounted slot terminals, but when the workspace is on the feed tab every slot pane is hidden with `display:none` (`src/pages/MissionWorkspace.tsx:849`). For a slot tab that has never been activated, `RunnerTerminal.measure()` sees a 0x0 rect and returns null for the constructor default 80x24 (`src/components/RunnerTerminal.tsx:765`). That leaves `sharedDims` null, so `session.resume(..., null, null)` still hits the backend 80x24 fallback. This is the exact common path after opening a mission, staying on feed, stopping/pausing, then clicking Resume from the feed overlay. The fallback needs to come from a visible workspace/container measurement or some cached last known workspace fit, not only from hidden terminal handles.

Verification:
- `pnpm run lint` passed.
- `git diff --check origin/main...pr-191` passed.

yicheng47:
Follow-up review finding:

1. Medium: reopening a stopped direct chat can wipe its terminal scrollback. `src/pages/RunnerChat.tsx:1013` still passes `active={active && !resuming && !starting}` to `RunnerTerminal`, so a stopped/crashed direct session is active whenever its tab is selected even though `disabled` is true. This PR made the activation effect unconditionally clear TUI viewports before the resize dance (`src/components/RunnerTerminal.tsx:724`); for dead sessions the resize rejects and there is no agent left to repaint, leaving the dimmed Session ended surface with a blank canvas instead of the preserved scrollback promised by the `disabled` prop. `MissionWorkspace` already gates `active` with `!dead`, so direct chat should probably do the same or `RunnerTerminal` should skip the wipe/dance while `disabledRef.current` is true.

The previous feed-tab resume fallback issue looks fixed by the new pane-container measurement.

Verification:
- `pnpm run lint` passed.
- `git diff --check origin/main...pr-191` passed.

PR #198, opened 2026-06-01
feat(runner-form): model combobox + unified selector styling

yicheng47:
## What

Started as "make `default` the default in the model picker" and grew into a model selector + runner-form UX pass.

### Model selector
- **New runner modal** now has a **Model** field (previously model was only editable *after* creation, in the edit drawer).
- Both surfaces use a shared **editable combobox** (`ModelField`): type any model name, or pick a curated alias from the `▾`.
  - claude-code suggests `default` / `opus` / `sonnet` / `haiku` (version-stable aliases — they don't go stale).
  - codex suggests `default` only (no stable alias scheme; type full names like `gpt-5-codex`).
  - Empty field = the runtime's own default. The backend passes `--model <name>` verbatim, so there's no closed set.

### Form polish
- **Field hints** render as an info-icon tooltip (new portal-based `Tooltip`, shows on hover **and** focus) instead of inline text that wrapped into the label. Removed low-value hints (Working directory, System prompt).
- **Unified dropdown styling** — `RuntimeSelect`, `StyledSelect` (Permission mode / Thinking effort), and the model combobox now share one look. (Heads-up: `StyledSelect` is shared, so Settings dropdowns inherit the restyle.)
- **Permission mode** is now a standard labeled field (description moved into the tooltip), and **defaults to Auto**.
- **Modal height** capped at `85vh` with a scrolling body, so tall modals scale with the window instead of overflowing off-screen (applies to all `Modal`s).

### Bug fixes
- **Outside-click now closes dropdowns inside modals/drawers** — the panels `stopPropagation` on bubbling `mousedown`, which swallowed the dropdowns' outside-click detection. Switched those listeners to the capture phase.
- **Selecting `default` / clearing the model now persists.** Pre-existing bug: the drawer sent JSON `null`, which deserializes into the backend's `Option<Option<String>>` as the outer `None` = "leave unchanged", so a pinned model could never be cleared. Now sends the empty string the backend already expects (it trims `""` → NULL).

## Notes
- No backend changes — all frontend.
- We dug into how Zed sources its (dynamic, ACP-reported) model lists and decided that's out of scope here; runner's terminal/flag integration can't replicate it without adopting ACP.

## Test plan
- `npm run lint` ✅
- `npm run build` (tsc && vite build) ✅
- Manual: create/edit runner, pick `default`/alias/custom, save and verify; outside-click + click-toggle on all selectors; tall modal scrolls.

### DIFF
--- a/src-tauri/src/commands/runner.rs
+++ b/src-tauri/src/commands/runner.rs
@@ -44,27 +44,19 @@ pub struct CreateRunnerInput {
     /// Permission mode the runner-edit form's dropdown chose. Mapped
     /// to concrete flags on the row's `args` column at create time
     /// via `router::runtime::apply_permission_mode`. Defaults to
-    /// `AcceptEdits`:
-    /// - claude-code → `--permission-mode acceptEdits`. Works on
-    ///   every plan and avoids the consent dialog.
-    /// - codex → no flag (codex has no edits-only middle on the
-    ///   wire; `AcceptEdits` is treated as `Default` there).
-    ///
-    /// `Auto` is opt-in for claude-code because the real `auto`
-    /// mode is plan/model-gated (Max/Team/Enterprise/API + a
-    /// supported model). `Bypass` is opt-in because claude-code
-    /// shows a one-time consent dialog the first time per user
-    /// account. Hidden in the form for runtimes without a
-    /// permission concept (shell / unknown).
+    /// `Auto`. The new-runner form defaults to codex, where Auto
+    /// maps to `--ask-for-approval on-request --sandbox workspace-write`.
+    /// For claude-code, Auto is still plan/model-gated; callers that
+    /// default to claude-code should send AcceptEdits explicitly.
     #[serde(default = "default_permission_mode")]
     pub permission_mode: crate::router::runtime::PermissionMode,
 }
 
-/// Default permission mode for new runners — `AcceptEdits`. Matches
-/// the frontend's dropdown default and the seed's runner args.
+/// Default permission mode for new runners — `Auto`. Matches the
+/// frontend's dropdown default and the seed's runner args.
 /// Pulled out so serde's `#[serde(default = "...")]` can name it.
 fn default_permission_mode() -> crate::router::runtime::PermissionMode {
-    crate::router::runtime::PermissionMode::AcceptEdits
+    crate::router::runtime::PermissionMode::Auto
 }
 
 // `handle` is intentionally excluded from updates: per arch §2.2 and §5.2
--- a/src-tauri/src/db.rs
+++ b/src-tauri/src/db.rs
@@ -117,14 +117,12 @@ const SEED_IMPL_RUNNER_ID: &str = "01K000DEFAULT000RUNNERIMPL01";
 const SEED_REVIEWER_RUNNER_ID: &str = "01K000DEFAULT000RUNNERREVW01";
 const SEED_TIMESTAMP: &str = "2026-05-03T00:00:00Z";
 
-// Auto permission mode args — `claude --permission-mode acceptEdits`.
-// Auto-approves edits but still asks for blast-radius operations
-// (shell commands outside the workspace, network calls). Crucially
-// does NOT trigger claude-code's one-time bypass-permissions consent
-// dialog, which used to break first-mission injection. Users can
-// flip individual runners to Bypass via the runner-edit form's
-// segmented control if they want the more aggressive default.
-const SEED_RUNNER_ARGS_JSON: &str = r#"["--permission-mode","acceptEdits"]"#;
+// Auto permission mode args for the default Codex seed:
+// `codex --ask-for-approval on-request --sandbox workspace-write`.
+// This matches the new-runner form's default runtime + permission
+// mode without relying on claude-code's plan-gated Auto mode.
+const SEED_RUNNER_ARGS_JSON: &str =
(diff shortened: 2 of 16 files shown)


yicheng47:
Review findings:

1. Medium: dropdowns can be clipped inside the new modal scroll body. In `src/components/ui/Overlay.tsx:57`, modal content is now `overflow-y-auto`, while `ModelField` and `StyledSelect` render absolute listboxes inside that clipped subtree. In New Runner, the Model/Permission dropdowns can extend behind the footer and the missing options are not reachable by scrolling. Portal/fixed-position these popovers like `Tooltip`, or render them outside the scroll container.

2. Medium: `StyledSelect` dropdowns in Settings can clip horizontally. In `src/components/ui/StyledSelect.tsx:115`, the menu changed to `left-0 w-full min-w-[240px]`, but Settings rows keep a 160px right-aligned trigger inside an overflow-clipped panel. The extra width extends past the right edge. Keep right alignment for default/narrow selects, add an align prop, or portal the menu.

Verification:
- `pnpm run lint` passed.
- `pnpm run build` passed with the existing large chunk warning.

yicheng47:
Both findings fixed in ea31924 by portaling the dropdown menus, as suggested.

Extracted a shared **`PopoverMenu`** (portals to `<body>`, fixed-positioned from the anchor's rect, repositions on scroll/resize, flips above when there's no room below, and clamps on-screen; outside-click checks the anchor **or** the menu so option clicks still register). Moved `RuntimeSelect`, `StyledSelect`, and the model combobox onto it.

1. **Modal scroll clip** — menus now render outside the `overflow-y-auto` body, so Model/Permission options no longer hide behind the footer.
2. **Settings horizontal clip** — the menu is no longer inside the clipped panel; it's clamped to the viewport, so the widened `min-width` can't spill past the edge.

Re-verified: `tsc --noEmit`, `eslint`, and `vite build` all pass.

(Also on this branch: f72bb65 makes terminal links require Cmd/Ctrl+click — standard terminal behaviour — which is a separate ask from this session; happy to split it into its own PR if you'd prefer.)

yicheng47:
Review findings:

1. Medium: Settings dropdown options are now treated as outside clicks. In `src/components/ui/StyledSelect.tsx:95`, menus render through `PopoverMenu`, which portals them to `document.body`; `src/components/SettingsModal.tsx:167` closes the modal on any document `mousedown` whose target is outside `cardRef`. Clicking a portaled menu option is outside `cardRef`, so Settings can close before the option click commits. Include the popover node in the inside-click check, or have the portaled menu stop the native mousedown from reaching Settings’ document listener.

2. Low: field hint tooltips do not actually show on focus in current usage. In `src/components/ui/Field.tsx:26`, `Tooltip` wraps a plain non-focusable `span`, so `Tooltip`’s focus handlers are unreachable by keyboard users. Add `tabIndex={0}` or use a button-like info trigger with appropriate styling.

Verification:
- `pnpm run lint` passed.
- `pnpm run build` passed with the existing large chunk warning.

The previous modal clipping findings look addressed by the new portaled `PopoverMenu`.

yicheng47:
Both fixed in d2d803d.

1. **Settings closing on option-select** — root cause confirmed: `SettingsModal` closes on a *native* `document` mousedown outside `cardRef`, and `PopoverMenu` portals the menu outside that card in the DOM. (`Modal`/`Drawer` were unaffected — they close via React `onMouseDown`, which follows React-tree bubbling through the portal, so the panel's `stopPropagation` already covers them.) Fix: `PopoverMenu` now attaches a native `mousedown` listener on the menu node that `stopPropagation`s, so clicks inside the portaled menu never reach any document/window outside-click listener. Reliable regardless of React's portal-event semantics, and protects any future native-listener consumer too.
2. **Hint tooltip not keyboard-focusable** — the info-icon trigger now has `tabIndex={0}` + `role="img"` (with a `focus-visible` ring), so `Tooltip`'s focus handlers fire on keyboard focus.

Re-verified: `tsc --noEmit`, `eslint`, `vite build` all pass.

Also pushed some unrelated UX tweaks from this session in 99ca155 (trimmed tooltips, working-directory path now wraps instead of truncating).

yicheng47:
Review findings:

1. Medium: New `claude-code` runners now default to a gated mode. In `src/components/CreateRunnerModal.tsx:55`, permission mode initializes to `auto`, and the create payload always sends it for supported runtimes at `src/components/CreateRunnerModal.tsx:122`. Since the default runtime is `claude-code`, new runners get `--permission-mode auto` by default. The app’s own copy says that mode requires Max/Team/Enterprise/API plus supported models and is not available on Pro at `src/components/ui/runtimes.ts:176`. That makes the default create path produce runners that can fail to start for Pro users. Keep `accept_edits` as the claude-code default, or make the default runtime-specific.

The previous Settings/dropdown and tooltip-focus findings look fixed.

Verification:
- `pnpm run lint` passed.
- `pnpm run build` passed with the existing large chunk warning.
- `git diff --check origin/main...origin/feat/model-default-placeholder` passed.

yicheng47:
Follow-up on the default-mode finding:

Preferred fix direction: make the default runtime `codex` and keep the create-form permission mode default as `auto`, so the default new-runner path produces Codex args `--ask-for-approval on-request --sandbox workspace-write` instead of gated Claude Code `--permission-mode auto`. The first-launch Build squad seed should move to `codex` as well, including its command and seeded permission args, so seeded runners match the new default behavior.

PR #202, opened 2026-06-18
feat(chat): add direct runtime chats

### LINKED ISSUE #195: feat: direct chat with runtime picker — skip the runner template
Spec: [`docs/features/25-direct-chat-runtime-picker.md`](./docs/features/25-direct-chat-runtime-picker.md)

## Motivation

The Start Chat modal today requires picking a **runner** — a pre-configured template (runtime + system prompt + model + effort). That's overhead for the most common use case: "I want to talk to claude-code in this directory." Users who haven't set up runners, or who want a throwaway chat, are forced through the template workflow first.

## Scope summary

- **Two-mode StartChatModal** — segmented control: "Runner" (existing flow) or "Runtime" (new: runtime dropdown + cwd picker, two fields, go).
- **`session_start_runtime` command** — spawns a direct PTY from a runtime name + cwd without a runner row. Ephemeral config, no template pollution.
- **Nullable `runner_id` on sessions** — runtime-spawned chats have no runner template. Sidebar/header show the runtime display name instead of `@<handle>`.
- **Hardcoded runtime registry** — `runtime_list` command returns known runtimes (`claude-code`, `codex`). Extensible later.

### Out of scope

- Model/effort/system prompt overrides in runtime mode (keep it simple, don't rebuild the runner form).
- User-defined runtimes, auto-detect installed CLIs.

See the spec for key decisions (two-mode vs merged form, nullable FK vs synthetic runner row, hardcoded registry vs plugin discovery).

yicheng47:
Summary
- Add direct runtime chat sessions persisted with agent runtime and command metadata.
- Add the direct/runtime chat picker plus Chat settings for default runtime args.
- Wire direct chat resume/archive/sidebar flows and clean up sidebar selection/loading states.

Closes #195

Validation
- pnpm exec tsc --noEmit
- pnpm run lint
- cargo check
- cargo test --workspace
- cargo fmt --check
- git diff --check

### DIFF
--- a/docs/impls/0013-direct-chat-runtime-picker.md
+++ b/docs/impls/0013-direct-chat-runtime-picker.md
@@ -1,0 +1,196 @@
+# Direct Chat Runtime Picker
+
+> Implements [#195](https://github.com/yicheng47/runner/issues/195). Product spec: [docs/features/25-direct-chat-runtime-picker.md](../features/25-direct-chat-runtime-picker.md).
+
+## Context
+
+Direct chats are currently runner-template chats end to end. `StartChatModal` fetches runners, `session_start_direct` requires a `runner_id`, `sessions.runner_id` is `NOT NULL`, direct-chat listing joins `runners`, and `RunnerChat` is routed through `/runners/:handle/chat/:sessionId` so it can fetch the runner by URL handle for the header and terminal runtime.
+
+Issue #195 adds the second direct-chat path: pick an agent runtime plus cwd, spawn immediately, and do not create a runner template. The main design constraint is that a runtime-only chat has no stable runner handle or runner row, so this cannot be just a modal change.
+
+## Decisions
+
+1. Replace the runner-scoped direct-chat route with `/chats/:sessionId`.
+2. Make `sessions.runner_id` nullable while preserving `ON DELETE CASCADE` for rows that still reference a runner.
+3. Add explicit agent-runtime metadata on `sessions` instead of reusing `sessions.runtime`.
+4. Keep runtime mode intentionally bare: runtime dropdown, cwd picker, Start. No model, effort, permission, args, env, or prompt fields.
+5. Runtime-only direct chats are resumable. Resume reconstructs the same ephemeral runner config from the session row instead of looking up `runners`.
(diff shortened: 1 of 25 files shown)


yicheng47:
Review findings:

1. High: reloading or directly opening a running chat renders it as stopped. In `src/pages/RunnerChat.tsx:583`, attach falls back to `"stopped"` when `location.state` is missing, even though `chatMeta.status` has already loaded from SQLite. The sync effect at `src/pages/RunnerChat.tsx:353` no-ops before the pane exists, so it never corrects the newly inserted pane. A `/chats/:sessionId` reload can show the paused overlay and disable `RunnerTerminal` for a live PTY. Use `state?.sessionStatus ?? chatMeta.status` or upsert from `chatMeta`.

2. Medium: runtime-only start/resume has no reliable sidebar refresh signal. The sidebar only refreshes direct chats on `session/exit`, `runner/activity`, `session/archived`, and `session/updated` in `src/components/Sidebar.tsx:296`. Runtime-only resume suppresses activity at `src-tauri/src/session/manager.rs:1802`, and `session_start_runtime` returns without emitting any session lifecycle event in `src-tauri/src/commands/session.rs:645`. Result: resumed runtime-only rows can stay dim/stopped in the sidebar until exit. Emit `session/updated` or a dedicated start/resume event after runtime-only status changes.

Verification:
- `pnpm run lint` passed.
- `cargo fmt --check` passed.

yicheng47:
Review findings:

1. Medium: stale navigation state can still override DB truth in `src/pages/RunnerChat.tsx:583`. Attach is now gated on loaded `chatMeta`, but it still seeds status with `state?.sessionStatus ?? chatMeta.status`. If a user opens a stopped chat, resumes it, navigates away, then goes back to that same history entry, the old `location.state.sessionStatus = "stopped"` wins over `chatMeta.status = "running"`. Because the sync effect ran before the pane existed and `startedKeyRef` blocks reattach, the UI can leave a live PTY disabled behind the stopped overlay. Since attach already waits for `chatMeta`, use `chatMeta.status` as the source of truth.

Verification:
- `pnpm run lint` passed.
- `cargo fmt --check` passed.

The prior two findings look fixed.

PR #203, opened 2026-06-18
fix(terminal): refit active pane after wake

### LINKED ISSUE #200: bug: terminal history stuck at stale width after laptop sleep/wake
## Description
After the laptop sleeps and is reopened, the terminal pane's history stays wrapped at a stale (narrow) width even though the window is wide — content occupies only the left portion of the pane and never reflows to the actual window width.

## Expected behavior
On wake, the active terminal should refit to the current container size (fit + PTY resize + redraw), so the live screen and subsequent output use the full pane width.

## Steps to reproduce
1. Open a chat/mission with an active terminal session.
2. Let the laptop go to sleep (lid close or idle).
3. Wake the laptop and reopen the app window.
4. Terminal content remains wrapped at the pre-sleep (or transient) narrow width.

## Relevant code
- `src/components/RunnerTerminal.tsx` — resize wiring listens to `window resize` (line ~440), a container `ResizeObserver` (~448), `window focus` and `document visibilitychange` (~472-473). None of these reliably fire on macOS wake-from-sleep in WKWebView, so `refitAndPush()` never runs.
- `src/components/RunnerTerminal.tsx:650-761` — the tab-activation effect (fit + atlas clear + SIGWINCH row-perturbation dance from #177) only runs on `active`/`sessionId` dep changes; the active tab doesn't change across sleep, so it never re-runs.
- `src-tauri/src/session/pty_runtime.rs:250-262` and `src-tauri/src/commands/session.rs:136-143` — PTY resize path (works; it's just never invoked on wake).
- `docs/impls/0012-force-redraw-on-tab-activation.md` — explains the SIGWINCH dance and why scrollback with hard-wrapped newlines can't reflow after the fact, which is why catching the wake event promptly matters.

Likely fix direction: hook a wake/focus signal that actually fires in Tauri on macOS (e.g. Tauri window focus event via `@tauri-apps/api/window`, or an OS resume notification from the Rust side) and run the same refit + redraw sequence as tab activation for the active terminal.

## Environment
- OS: macOS (Darwin 24.6.0)
- Version: 0.1.21

yicheng47:
Summary
- Emit an app/resumed event from Tauri when the OS resumes the app.
- Reuse the active terminal refit/redraw path for app resume, native window focus, browser focus, and visibility restore.
- Run a delayed second wake pass so WKWebView layout can settle before the final PTY resize.

Closes #200

Validation
- pnpm exec tsc --noEmit
- pnpm run lint
- cargo check
- cargo fmt --check
- git diff --check

Manual
- Sleep/wake smoke looked fine per local testing.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -15,7 +15,7 @@ use std::sync::Arc;
 #[cfg(target_os = "macos")]
 use tauri::menu::{AboutMetadataBuilder, PredefinedMenuItem};
 use tauri::menu::{Menu, MenuBuilder, MenuItemBuilder, SubmenuBuilder};
-use tauri::{AppHandle, Manager, Wry};
+use tauri::{AppHandle, Emitter, Manager, Wry};
 use tauri_plugin_log::{Builder as LogBuilder, RotationStrategy, Target, TargetKind};
 
 /// Bundle identifier as declared in `tauri.conf.json`. Used by:
@@ -297,8 +297,14 @@ pub fn run() {
             // runtime this still demotes DB rows cleanly. We
             // tolerate kill failures: best-effort path, and the
             // startup cleanup is the safety net on next launch.
-            if let tauri::RunEvent::ExitRequested { .. } = event {
-                stop_running_sessions_on_quit(app_handle);
+            match event {
+                tauri::RunEvent::ExitRequested { .. } => {
+                    stop_running_sessions_on_quit(app_handle);
+                }
+                tauri::RunEvent::Resumed => {
+                    let _ = app_handle.emit("app/resumed", ());
+                }
+                _ => {}
             }
         });
 }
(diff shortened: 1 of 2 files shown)


yicheng47:
Review findings:

1. Low: `src/components/RunnerTerminal.tsx:580` calls `getCurrentWindow()` outside a `try/catch`. In browser dev preview, `getCurrentWindow()` synchronously dereferences `window.__TAURI_INTERNALS__`, so the `.catch()` on `onFocusChanged()` never runs and the effect throws before cleanup is returned. Existing Tauri API usage guards this path with `try/catch`, e.g. `src/lib/appZoom.ts:17`. Wrap the native focus listener registration in `try/catch` or guard for `window.__TAURI_INTERNALS__`.

Verification:
- `pnpm run lint` passed.
- `cargo check` passed.
- `cargo fmt --check` passed.

I did not run the manual sleep/wake smoke test.

PR #208, opened 2026-06-19
fix(terminal): keep mission PTY panes measurable

yicheng47:
Summary
- Keep inactive mission PTY panes invisible but measurable instead of display:none so hidden xterms can fit and replay initial Codex frames.
- Let RunnerTerminal drain replay once the pane is measurable, even before the tab is active.
- Avoid focus/display wake clear+SIGWINCH to remove terminal flash.

Validation
- pnpm exec tsc --noEmit
- pnpm run lint
- git diff --check
- Manual: user confirmed fresh mission Codex input background now matches expected behavior.

### DIFF
--- a/src/components/RunnerTerminal.tsx
+++ b/src/components/RunnerTerminal.tsx
@@ -214,9 +214,11 @@ export const RunnerTerminal = forwardRef<
     ({
       focus = false,
       forceResizeDance = false,
+      pushBackendSize = false,
     }: {
       focus?: boolean;
       forceResizeDance?: boolean;
+      pushBackendSize?: boolean;
     } = {}) => {
       if (!activeRef.current) return;
       const t = termRef.current;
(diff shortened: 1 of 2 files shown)


yicheng47:
Review findings:

1. Medium: wake/focus can leave xterm and the backend PTY at different sizes. In `src/components/RunnerTerminal.tsx:546`, the wake path now calls `refreshActiveTerminal()` without a resize push. That function still runs `fit.fit()`, but returns before `api.session.resize(...)` unless `forceResizeDance` is true. If macOS wake/display changes skip `ResizeObserver` as the old comment warned, the frontend grid updates while the PTY keeps the stale winsize, so future TUI output can drift/wrap incorrectly. Keep the no-flash behavior, but still push changed cols/rows on wake.

Verification:
- `pnpm run lint` passed.

yicheng47:
Addressed the review finding.

Change:
- Wake/focus refresh now distinguishes renderer refresh from forced redraw. It still skips the clear + nudged SIGWINCH path, but it pushes changed cols/rows to the backend when `fit.fit()` changes the xterm grid.

Validation:
- `pnpm exec tsc --noEmit`
- `pnpm run lint`
- `git diff --check`

PR #212, opened 2026-06-20
Fix Codex mission spawn plumbing

yicheng47:
Summary:
- Keep Codex mission first-turn delivery deterministic via spawn-time positional argv, while still granting mission bus access with --add-dir.
- Pass estimated mission terminal geometry into mission_start and preserve it through backend spawn registration so mission PTYs do not start at 80x24.
- Defer mission terminal snapshot replay until panes are active/measurable, then avoid clearing freshly replayed frames during activation.
- Keep inactive mission panes hidden while preserving mounted xterm state and refitting on activation.

Validation:
- pnpm exec tsc --noEmit
- pnpm run lint
- git diff --check
- cargo test --workspace

Known follow-up:
- Codex mission prompt is now sent, but the prompt/input background can still render black on fresh mission start. That rendering issue remains for follow-up investigation.

### DIFF
--- a/src-tauri/src/commands/mission.rs
+++ b/src-tauri/src/commands/mission.rs
@@ -492,6 +492,15 @@ pub(crate) async fn mission_start_impl(
     app: &tauri::AppHandle,
     input: StartMissionInput,
 ) -> Result<StartMissionOutput> {
+    mission_start_impl_with_size(state, app, input, None).await
+}
+
+async fn mission_start_impl_with_size(
+    state: &AppState,
+    app: &tauri::AppHandle,
+    input: StartMissionInput,
+    initial_size: Option<(u16, u16)>,
+) -> Result<StartMissionOutput> {
     use crate::event_bus::{BusEmitter, TauriBusEvents};
     use crate::router::{
         open_log_for_mission, CompositeBusEmitter, Router, RouterSubscriber, StdinInjector,
@@ -678,6 +687,7 @@ pub(crate) async fn mission_start_impl(
             events_log_path.clone(),
             state.db.clone(),
             first_turn,
+            initial_size,
         );
         match register_res {
             Ok(pending) => {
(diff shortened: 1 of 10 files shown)


yicheng47:
Review findings:

1. Medium: stopped mission tabs can never replay their terminal snapshot after reload / first open. `src/components/RunnerTerminal.tsx:729` now refuses to drain `outputSnapshot` unless `activeRef.current` is true, but mission slot panes pass `active={active && !dead && !resuming && !starting}` at `src/pages/MissionWorkspace.tsx:1235`. For a stopped/crashed slot, `dead` keeps `active` false forever even while its tab is visible; the comments/state at `src/pages/MissionWorkspace.tsx:1214` still expect dimmed scrollback behind the paused card. The result is a blank terminal under “Mission paused” after reopening a paused mission or opening a slot that died before it was viewed. Split “visible/measurable” from “input-active”, or let replay drain when the container is visible even if the terminal is disabled.

2. Medium: inactive mission tabs now buffer live output without a cap until first activation. `src/pages/MissionWorkspace.tsx:1267` hides inactive panes with `hidden`, and the new `activeRef` gate at `src/components/RunnerTerminal.tsx:729` means `replayDone` remains false for every background tab. While the user stays on the feed or another tab, each `session/output` event is appended to `pendingLiveRef` at `src/components/RunnerTerminal.tsx:791` with no ring limit. A noisy worker can grow webview memory for the life of the workspace and then force a large replay burst when the tab is opened. Either cap this queue to the same policy as the backend output buffer, or refetch/drain from a bounded snapshot on activation instead of retaining every live chunk in React memory.

Verification:
- `pnpm run lint` passed.
- `git diff --check origin/main...origin/pr/212` passed.

yicheng47:
Review follow-up:

No remaining issues found in the updated head. The new `Fix mission terminal replay lifecycle` commit addresses the prior findings:
- stopped mission slots now stay replay-active while visible but disabled for input;
- hidden tabs cap pending live output and refresh from the backend snapshot ring before replay after overflow.

Verification:
- `pnpm run lint` passed.
- `git diff --check origin/main...origin/pr/212` passed.

PR #214, opened 2026-06-20
fix(session): tighten Codex rollout capture

yicheng47:
## Summary
- capture Codex rollout timestamps from the actual PTY spawn point instead of row registration
- make rollout matching choose the earliest unclaimed matching session_meta after that spawn time
- reject rollout metadata older than the spawn timestamp and add parser regression tests

## Validation
- cargo test -p runner codex_capture
- cargo test -p runner session::manager
- cargo test -p runner
- git diff --check

### DIFF
--- a/src-tauri/src/session/codex_capture.rs
+++ b/src-tauri/src/session/codex_capture.rs
@@ -23,7 +23,7 @@
 
 use std::collections::HashSet;
 use std::io::{BufRead, BufReader};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex, OnceLock};
 use std::time::{Duration, Instant};
 
@@ -34,15 +34,6 @@ use crate::db::DbPool;
 
 const CAPTURE_TIMEOUT_SECS: u64 = 30;
 const POLL_INTERVAL_MS: u64 = 400;
-/// Slack between the row's `started_at` and the rollout's
-/// `payload.timestamp`. Codex stamps `payload.timestamp` slightly
-/// AFTER our spawn returns, so under normal conditions the rollout
-/// is always >= our `started_at`. We allow a small negative window
-/// (1s) to absorb minor clock skew between the spawn-time
-/// `Utc::now()` call and codex's. A wider window risks two chats
-/// started in the same cwd within seconds of each other capturing
-/// each other's id — see `claimed_rollouts` below.
-const TIMESTAMP_SLACK_MS: i64 = 1000;
 
 /// Process-shared set of rollout paths that some watcher has already
 /// captured. Two concurrent watchers polling the same date dir might
@@ -103,6 +94,7 @@ fn run(session_id: String, spawn_cwd: String, started_at: DateTime<Utc>, pool: A
                 Ok(e) => e,
                 Err(_) => continue,
             };
+            let mut matches: Vec<(DateTime<Utc>, PathBuf, String)> = Vec::new();
             for entry in entries.flatten() {
                 let path = entry.path();
                 if done.contains(&path) {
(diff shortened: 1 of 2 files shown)


yicheng47:
## Review — solid fix, ship it

Core change is correct and well-reasoned; tests + clippy pass locally (`cargo test -p runner codex_capture`, `cargo clippy -p runner --lib`). One residual race worth knowing about (pre-existing, not introduced here) and a minor test-coverage gap. Nothing blocking.

### Why this is the right shape

The three changes reinforce each other and fix the sibling-fusion bug. With an exact per-watcher floor + earliest selection, two chats started seconds apart in the same cwd pair up correctly:

- Watcher A (floor `T0`) sees rollouts A and B, picks the earliest → A. ✓
- Watcher B (floor `T1 > T0`) sees only B (A rejected as `ts_A < T1`), picks B. ✓

The old `1s` slack is what *caused* the fusion: it let watcher B match A's rollout (`ts_A + 1000ms >= T1`), and arbitrary `read_dir` order made which one won nondeterministic. Moving the floor to the real spawn instant also makes the slack unnecessary — codex's `payload.timestamp` is stamped by a process that doesn't exist until after `runtime.spawn` returns, so on a single machine it's strictly later than `spawn_started_at_dt`.

### Issues / risks

- **Residual race (not introduced here):** if chat A spawns first but its rollout file is written *slower* than later sibling B's, watcher A (floor `T0`) sees only B, treats it as the earliest match, and claims B's id — fusing in the other direction. `claimed_rollouts` prevents two watchers grabbing the *same* file but not a watcher grabbing the *wrong* file. Fundamental to id-by-timestamp matching with no caller-provided id; this PR makes it strictly rarer, never worse. Fine to accept.
- **Slack removal trades one rare failure for another:** an NTP step-backward in the sub-second window between our `Utc::now()` and codex's could push `ts < started_after` and make us reject our own rollout. Degrades gracefully (NULL key → fresh spawn, same as today) and is far less likely than the fusion bug it eliminates. Good trade.

### Suggestions

1. **Drop a one-line comment near the `matches.sort_by`** noting that earliest-selection assumes rollout write order tracks spawn order — so the residual race above is documented at the spot it lives.
2. **Lift the selection step into a pure helper and unit-test it.** The genuinely new behavior (sort + earliest-unclaimed) is the logic most likely to regress and is currently untested — `run()` is hard to test (threads + filesystem + `$HOME`), but the selection is easy to isolate:

```rust
/// Earliest unclaimed match wins. Assumes codex writes rollout files
/// in spawn order; a sibling spawned later but flushed first can be
/// mis-claimed (best-effort, degrades to a fresh spawn).
fn select_earliest_unclaimed(
    mut matches: Vec<(DateTime<Utc>, PathBuf, String)>,
    is_claimed: impl Fn(&PathBuf) -> bool,
) -> Option<(PathBuf, String)> {
    matches.sort_by(|a, b| a.0.cmp(&b.0).then_with(|| a.1.cmp(&b.1)));
    matches
        .into_iter()
        .find(|(_, path, _)| !is_claimed(path))
        .map(|(_, path, id)| (path, id))
}
```

Then a test pinning the sibling-pairing guarantee:

```rust
#[test]
fn select_earliest_unclaimed_picks_earliest_then_skips_claimed() {
    let t0 = "2026-06-20T12:00:00Z".parse::<DateTime<Utc>>().unwrap();
    let early = (t0, PathBuf::from("rollout-early.jsonl"), "id-a".to_string());
    let late = (t0 + chrono::Duration::seconds(2), PathBuf::from("rollout-late.jsonl"), "id-b".to_string());

    // Earliest wins.
    let got = select_earliest_unclaimed(vec![late.clone(), early.clone()], |_| false);
    assert_eq!(got, Some((early.1.clone(), "id-a".to_string())));

    // Earliest already claimed → fall through to the next.
    let got = select_earliest_unclaimed(vec![late.clone(), early.clone()], |p| p == &early.1);
    assert_eq!(got, Some((late.1.clone(), "id-b".to_string())));
}
```

Both are optional — coverage today is fine for the parser, this just locks in the sibling-pairing behavior that is the whole point of the PR.

### Nits

- `started_at_dt` at `manager.rs:817`, `:1242`, `:1717` is now used only for `.to_rfc3339()`. Harmless; keeping the binding preserves symmetry across the three sites.
- `parse_session_meta` switching to `&Path` is the right idiomatic move (satisfies `ptr_arg`).

PR #225, opened 2026-06-23
Fix mission terminal bracketed paste snapshot replay

### LINKED ISSUE #223: bug: mission terminal multiline paste submits each line
## Description
Pasting a multiline code snippet into a mission session terminal can fail to preserve the paste as one bracketed-paste input. The newline characters are treated like raw Enter submissions, so the agent receives each pasted line as a separate message/turn.

Observed with a Go function pasted into the mission slot PTY. The same failure was reported as specific to the mission session terminal.

Likely cause from code investigation: `RunnerTerminal` resets xterm before snapshot replay, and backend snapshot restoration only rehydrates alt-screen state. It does not track or synthesize bracketed paste mode (`\x1b[?2004h` / `\x1b[?2004l`), so after a mission pane reattach/replay xterm can believe bracketed paste is disabled while the underlying agent TUI expects it.

## Expected behavior
Pasting multiline text into a mission session terminal should be delivered as one paste block, preserving embedded newlines inside the agent input editor and submitting only when the user explicitly presses Enter.

## Steps to reproduce
1. Start a mission with a live runner session.
2. Open the runner's PTY tab in the mission workspace, especially after the session has already emitted startup output and the pane replays a snapshot.
3. Paste a multiline code snippet, for example a Go function with comments and blank lines.
4. Observe that each line is sent/submitted as a separate message instead of remaining in one input block.

## Relevant code
- `src/components/RunnerTerminal.tsx`: snapshot replay calls `t.reset()` before replaying buffered output, which clears xterm terminal modes including bracketed paste state.
- `src-tauri/src/session/manager/mod.rs`: `SessionState` tracks `alt_screen_on` only; no bracketed-paste state is stored.
- `src-tauri/src/session/manager/output.rs`: `output_snapshot` prepends synthetic `\x1b[?1049h` when `alt_screen_on` is true, but has no equivalent for `\x1b[?2004h`.
- `src-tauri/src/session/manager/output.rs`: `inject_stdin` treats xterm data as raw keystroke bytes, which is correct only if xterm preserved bracketed paste markers for multiline paste.

## Environment
- Version: 0.2.4

yicheng47:
Fixes #223.

Summary:
- Track bracketed paste mode from PTY output alongside alt-screen state.
- Synthesize bracketed paste enable in session snapshots after xterm reset/replay.
- Add focused scanner and snapshot replay tests.

Validation:
- cargo fmt
- cargo test -p runner bracketed_paste
- cargo test -p runner output_snapshot_prepends
- git diff --check
- Manual smoke test confirmed by Jason

### DIFF
--- a/src-tauri/src/session/manager/mod.rs
+++ b/src-tauri/src/session/manager/mod.rs
@@ -76,6 +76,22 @@ const CLAUDE_LAUNCH_GATE_GRACE: Duration = Duration::from_millis(1500);
 #[cfg(test)]
 const CLAUDE_LAUNCH_GATE_GRACE: Duration = Duration::from_millis(0);
 
+fn scan_mode_transition(bytes: &[u8], patterns: &[(&[u8], bool)]) -> Option<bool> {
+    let mut latest: Option<(usize, bool)> = None;
+    for (needle, state) in patterns {
+        if bytes.len() < needle.len() {
+            continue;
+        }
+        if let Some(pos) = bytes.windows(needle.len()).rposition(|w| w == *needle) {
+            latest = match latest {
+                Some((p, _)) if p >= pos => latest,
+                _ => Some((pos, *state)),
+            };
+        }
+    }
+    latest.map(|(_, state)| state)
+}
+
 /// Returns the resulting alt-screen state if `bytes` contains one or
 /// more enter/exit alt-screen escapes; `None` when no such escape is
 /// present. Recognized escapes: `\x1b[?1049h` / `\x1b[?1049l` (the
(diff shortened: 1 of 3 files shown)


yicheng47:
Review findings:

No must-fix findings. I reviewed the bracketed paste snapshot replay change in `src-tauri/src/session/manager/mod.rs`, `src-tauri/src/session/manager/output.rs`, and `src-tauri/src/session/manager/tests.rs` before merge. The implementation correctly tracks `?2004h` / `?2004l`, synthesizes the active mode during `output_snapshot`, and clears the mode state on purge/resume alongside alt-screen state.

Verification:
- `git diff --check` passed.
- `cargo fmt --check` passed.
- `cargo test -p runner scan_` passed.
- `cargo test -p runner output_snapshot` passed.

Follow-up check after merge: PR #225 is merged and CI reports Frontend and Backend success.

PR #231, opened 2026-06-24
Fix terminal repaint after app resume

### LINKED ISSUE #230: bug: mission terminal black screen after system wake
## Description
After the system wakes from sleep, the active mission terminal can render as a black screen. Switching away to another tab and back fixes the terminal immediately.

Likely cause from code investigation: the app emits `app/resumed` on Tauri `RunEvent::Resumed`, and `RunnerTerminal` handles it through `scheduleWakeRefit()`. That wake path calls `refreshActiveTerminal({ pushBackendSize: true })`, which clears the WebGL atlas and refreshes xterm, but if the terminal cols/rows are unchanged it hits the resize dedupe and does not force a PTY redraw.

Tab switching fixes the same terminal because the activation path calls `refreshActiveTerminal({ focus: true, forceResizeDance: true })`, which does the row-perturbing SIGWINCH dance and forces the agent TUI to repaint. The wake path should likely use the same stronger redraw path, possibly after layout has settled.

## Expected behavior
After macOS/system wake, the visible mission terminal should repaint itself without requiring a manual tab switch.

## Steps to reproduce
1. Open a mission workspace and focus a runner PTY tab.
2. Let the system sleep.
3. Wake the system and return to Runner.
4. Observe the mission terminal is black-screened.
5. Switch to another tab and back.
6. Observe the terminal repaints correctly.

## Relevant code
- `src-tauri/src/lib.rs`: `RunEvent::Resumed` emits `app/resumed` to the frontend.
- `src/components/RunnerTerminal.tsx`: `scheduleWakeRefit()` listens for focus/visibility/app-resumed events and calls `refreshActiveTerminal({ pushBackendSize: true })`.
- `src/components/RunnerTerminal.tsx`: `refreshActiveTerminal` dedupes same-size backend resize when `forceResizeDance` is false, so wake can skip the SIGWINCH repaint when geometry did not change.
- `src/components/RunnerTerminal.tsx`: the tab activation effect calls `refreshActiveTerminal({ focus: true, forceResizeDance: true })`, which performs the same redraw action that fixes the black screen.
- `docs/impls/archive/0012-force-redraw-on-tab-activation.md`: prior tab-activation black-canvas fix explains why the row-perturbing SIGWINCH dance is needed when xterm refresh alone cannot recover the rendered state.

## Environment
- Version: 0.2.4

yicheng47:
Fixes #230.\n\n## Summary\n- Make the terminal wake refresh path report whether it actually refreshed a measurable active pane.\n- Route Tauri app/resumed through the existing row-perturbing resize dance so visible TUI terminals get a forced SIGWINCH repaint after macOS wake.\n- Keep ordinary window focus / visibility wakes on the lighter local refresh path, with app-resume getting its own delayed retry window if layout is not ready immediately.\n\n## Validation\n- pnpm exec tsc --noEmit\n- pnpm run lint\n- git diff --check\n\nManual macOS sleep/wake verification not run.

### DIFF
--- a/src/components/RunnerTerminal.tsx
+++ b/src/components/RunnerTerminal.tsx
@@ -234,13 +234,13 @@ export const RunnerTerminal = forwardRef<
       forceResizeDance?: boolean;
       pushBackendSize?: boolean;
     } = {}) => {
-      if (!activeRef.current) return;
+      if (!activeRef.current) return false;
       const t = termRef.current;
       const fit = fitRef.current;
       const node = containerRef.current;
-      if (!t || !fit || !node) return;
+      if (!t || !fit || !node) return false;
       const rect = node.getBoundingClientRect();
-      if (rect.width <= 0 || rect.height <= 0) return;
+      if (rect.width <= 0 || rect.height <= 0) return false;
       try {
         fit.fit();
         tryDrainReplayRef.current?.();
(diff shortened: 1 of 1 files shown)


yicheng47:
Review findings:

No remaining must-fix issues found. The PR matches the fixed working-tree version I reviewed: app/resumed uses the forced row-perturbing resize path, and the retry/reset window now covers the earlier focus/visibility scheduling race.

Verification:
- `pnpm run lint` passed.

Residual risk: I did not run a manual macOS sleep/wake repro locally.

PR #232, opened 2026-06-24
Fix Codex session key capture

### LINKED ISSUE #229: bug: codex resume can bind to the wrong session
## Description
Codex resume can attach to the wrong Codex conversation/session. The resume argv path itself appears valid (`codex resume <uuid>`), but the persisted `sessions.agent_session_key` can be wrong before resume happens.

Likely cause from code investigation: Codex does not accept a caller-assigned session id on fresh spawn, so Runner captures Codex's generated id by polling `$HOME/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl` after spawn. The capture code matches rollout files by cwd + timestamp and then picks the earliest unclaimed matching rollout. For sibling Codex sessions in the same cwd, if rollout files are written/flushed out of spawn order, one watcher can claim the other session's rollout id and persist it into the wrong Runner session row. Later resume faithfully runs `codex resume <wrong uuid>`, so the visible PTY is bound to the wrong Codex conversation.

There is also a coverage gap: current tests cover that Codex resume skips first-turn injection and that mission resume grants `--add-dir`, but they do not assert that the resumed `SpawnSpec.args` includes the exact `agent_session_key` from the row being resumed.

## Expected behavior
Resuming a stopped Codex chat or mission slot should always pass that same Runner session row's native Codex session id to `codex resume <uuid>`, so the PTY restores the intended Codex conversation and not a sibling/newer session.

## Steps to reproduce
1. Start multiple Codex sessions in Runner with the same working directory, especially close together or in a multi-slot mission.
2. Let Codex create rollout files and let Runner capture `agent_session_key` values.
3. Stop one Codex session.
4. Resume it.
5. Observe that the resumed PTY can show or continue the wrong Codex conversation.

## Relevant code
- `src-tauri/src/session/codex_capture.rs`: `spawn_capture` polls Codex rollout files and writes `payload.id` into `sessions.agent_session_key`.
- `src-tauri/src/session/codex_capture.rs`: `select_earliest_unclaimed` explicitly assumes rollout files are written in spawn order; the comment notes that a later sibling flushing first can be mis-claimed.
- `src-tauri/src/session/manager/spawn.rs`: `SessionManager::resume` reads `agent_session_key` from the session row and passes it to `router::runtime::resume_plan`.
- `src-tauri/src/router/runtime.rs`: Codex resume plan builds `codex resume <uuid>` when the key is a UUID, so a wrong persisted key directly binds the resume to the wrong Codex session.
- `src-tauri/src/session/manager/tests.rs`: existing Codex resume tests do not assert exact resume argv binding to the row's key.

## Environment
- Version: 0.2.4

yicheng47:
Fixes #229.

Summary:
- Harden Codex native session-key capture so ambiguous same-cwd rollouts fail closed instead of persisting a sibling conversation id.
- Prefer pid-owned rollout detection, then a Codex-only Runner prompt marker for fresh first-turn spawns, then the guarded sole-owner fallback.
- Keep Claude Code on its native assigned-id path; the marker workaround is Codex-only.
- Retry Codex capture after first real input, emit session/updated on successful capture, and refresh direct chat plus mission session UI.
- Expose session_key in direct-chat runtime metadata and the mission runner rail, showing NULL when capture is unavailable.
- Add the implementation plan at docs/impls/0017-codex-session-key-capture.md.

Validation:
- cargo test -p runner codex_capture
- cargo test -p runner codex_mission_spawn_grants_event_log_dir_to_sandbox
- cargo test -p runner mission_spawn_worker_preamble_lands_as_trailing_positional_argv_with_brief
- cargo test -p runner codex_resume_skips_first_prompt_injection
- cargo test -p runner session_get_returns_agent_session_key_for_direct_chat
- cargo test -p runner session_list_returns_agent_session_key_for_mission_rows
- cargo fmt --check
- pnpm exec tsc --noEmit
- pnpm run lint
- cargo test --workspace
- git diff --check

Manual smoke:
- Started a multi-Codex mission after adding the Codex-only marker path and confirmed session_key populates instead of every mission runner staying NULL.

### DIFF
--- a/docs/impls/0017-codex-session-key-capture.md
+++ b/docs/impls/0017-codex-session-key-capture.md
@@ -1,0 +1,148 @@
+# Codex Session-Key Capture Hardening
+
+## Status
+
+Planned for issue #229.
+
+## Problem
+
+Codex resume can bind a Runner session row to the wrong native Codex conversation. The resume argv path is already correct: when `sessions.agent_session_key` is set, `router::runtime::resume_plan("codex", Some(key))` builds `codex resume <uuid>`. The weak point is the post-spawn capture of that key.
+
+Codex does not currently expose a fresh-spawn flag for caller-assigned session ids. Runner therefore watches `$CODEX_HOME/sessions/YYYY/MM/DD/rollout-*.jsonl`, reads the first `session_meta` row, and persists `payload.id` into `sessions.agent_session_key`. Today that watcher matches by cwd and spawn timestamp, then picks the earliest unclaimed rollout. That is not strong enough when several Codex sessions start close together in the same cwd: if rollout files flush out of spawn order, a watcher can claim a sibling session's native Codex id. The next resume faithfully resumes the wrong conversation.
+
+The safe invariant is: stale or missing resume is acceptable; wrong resume is not.
+
+Claude Code is intentionally separate: Runner can assign the native Claude session id at fresh spawn with `--session-id <uuid>` and later resume it with `--resume <uuid>`. The filesystem capture and prompt-marker workaround in this plan are Codex-only because Codex lacks the equivalent fresh-spawn id flag.
+
+## Goals
+
+- Persist a Codex `agent_session_key` only when capture is unambiguous for that Runner session row.
+- Keep `codex resume <uuid>` using the key from the exact row being resumed.
+- Fail closed: leave `agent_session_key = NULL` when capture cannot prove ownership.
+- Expose the captured key in the direct-chat side panel for debugging and operator confidence.
+
+## Non-Goals
+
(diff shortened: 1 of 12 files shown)


yicheng47:
Review findings:

No remaining issues found. The previous marker-race finding is fixed: marker-expected scans now stay pending when no marker is present, so they do not permanently fail through the ambiguous fallback path before Codex appends the marker-bearing prompt line.

Verification:
- `cargo test -p runner codex_capture` passed.
- `cargo test -p runner codex_mission_spawn_grants_event_log_dir_to_sandbox` passed.
- `cargo test -p runner mission_spawn_worker_preamble_lands_as_trailing_positional_argv_with_brief` passed.
- `cargo test -p runner codex_resume_skips_first_prompt_injection` passed.
- `cargo test -p runner session_get_returns_agent_session_key_for_direct_chat` passed.
- `cargo test -p runner session_list_returns_agent_session_key_for_mission_rows` passed.
- `pnpm run lint` passed.
- `cargo fmt --check` passed.
- `cargo test --workspace` passed: 326 tests across workspace crates/integration suites.

PR #251, opened 2026-07-03
Fix pane fill and preserve pane tabs

yicheng47:
## Summary
- Make sidebar new-chat spawns fill/focus the active tab's empty pane and inherit group pin before navigating.
- Add a freshness guard for just-assigned pane sessions so React Router's deferred navigation cannot let the vanished-session sweep evict a newly spawned pane fill.
- Preserve multiple pane tabs per window: creating panes from a non-member chat now adds another pane tab instead of replacing the previous one, and opening any member reactivates/focuses that member's pane tab.
- Add terminal resume/spawn repaint and dimension guards for the open cross-pane rendering investigation, keeping the intentional diagnostics.
- Align Window/Tab/Pane terminology in docs and remove the incorrect one-persistent-tab wording.

## Tests and checks
- `pnpm exec vitest run src/lib/paneLayout.test.ts`
- `pnpm exec vitest run`
- `pnpm exec tsc --noEmit`
- `pnpm run lint`
- `cargo fmt --check`
- `cargo test --workspace`
- `cargo clippy --workspace --all-targets`
- `git diff --check`
- GitHub CI: Frontend and Backend passed on the latest pushed commit.

## Manual smoke
- Create A+B, navigate to non-member C, create C+empty, then reopen A/B. Expected: A+B reactivates intact and C+empty remains its own pane tab.
- Click a chat that belongs to an inactive pane tab. Expected: that tab activates and the clicked chat's pane receives pane/keyboard focus.
- Use sidebar New Chat / Cmd+T while an active tab has an empty pane. Expected: the new chat fills the empty pane instead of opening as a separate single-pane tab.
- Manual tests passed in the pairing loop.

## Review
- Reviewer re-review is clean after the inactive-pane-tab sidebar focus fix.


yicheng47:
Review findings:

1. Medium: the immediate ownership report is missing for the sidebar row-pick fill path. In `src/components/Sidebar.tsx:827`, `openDirectChat` still calls `assignSessionToPane` for an existing non-member when the active group has a focused empty pane, then navigates to that session. That is the same explicit pane-membership change plus route transition that `reportSubjectsNow` was added to cover, and it is especially likely to involve a duplicated subject because the picked chat can already be open in another window. Until the debounced `useReportSubjects` write fires, the backend registry can still lack the newly visible pane subject for this window, so the ownership/overlay race remains for the “pick a chat from the sidebar” empty-pane path. Call `reportSubjectsNow(visibleSessionIds(getPaneLayout().root).map(...))` immediately after this `assignSessionToPane` as well, or factor the report into a helper shared by the three assignment sites.

Verification:
- `pnpm exec vitest run src/lib/windowFocus.test.ts src/lib/paneLayout.test.ts src/lib/groupPinning.test.ts` passed.
- `pnpm run lint` passed.
- `git diff --check HEAD^ HEAD` passed.

PR #252, opened 2026-07-04
Windows + WSL port (run agents in WSL, native host agents, i18n, multi-window fixes)

Lycheen97:
## What this is

A **Windows + WSL port** of Runner, branched from v0.2.10. Agents run inside
WSL by default; a per-runner **execution target** can instead run a
Windows-installed agent directly on the host. Also ships Chinese localization
and a batch of Windows-specific fixes (multi-window, terminal IME, shortcuts).

Verified end-to-end on Windows 11 + WSL2 (Ubuntu).

### WSL execution target (core)
- New `execution_target` column on runners (migration `0008`): `NULL`/`"wsl"`
  runs the agent inside a WSL distro via `wsl.exe`; `"native"` runs it on the
  Windows host.
- `pty_runtime` gains an injectable `CommandShaper`. The Windows build wires it
  to `session::wsl`, which shapes each spawn into `wsl.exe -d <distro>`,
  installs the embedded Linux `runner` CLI into the distro (so agents can emit
  signals onto the event bus), and holds spawned relays in a Job Object so they
  die with the app.
- Native target routes through `cmd.exe /c` so Windows agent shims
  (`codex.cmd`, `claude`) resolve via `PATHEXT`.

### Chinese i18n
- `LangProvider` + `t()` with the English source string as the key (missing
  translations fall back to English, so wrapping is always safe). Live language
  switch in Settings; default zh.

### Windows-specific fixes
- **Multi-window**: build secondary windows off the event-loop thread (async
  command + spawned thread) — a synchronous `build()` deadlocks on WebView2 and
  yields a blank, unclosable window (tauri #13963). Ctrl+N is bound in the
  frontend because the menu accelerator is swallowed by WebView2 on Windows
  (tauri #6365).
- **Terminal**: disable the WebGL renderer (its D3D surface makes GameDVR
  mis-detect the app as a game and pop `ms-gamingoverlay`), and re-anchor
  xterm's helper textarea so the CJK IME candidate window tracks the cursor in
  WebView2.
- **Platform-aware shortcuts** (⌘ ↔ Ctrl) for tab-switch and the sidebar
  toggle.
- MCP unix-socket transport gated behind `#[cfg(unix)]`; window-close on
  Windows/Linux exits the app (runs the agent-cleanup path) instead of hiding.

### Known limitations (intentionally deferred)
- MCP server disabled on the Windows host (agents reach MCP from inside WSL;
  core mission coordination flows through the NDJSON event log, not MCP).
- WSL distro hardcoded to `"Ubuntu"` — no detection / Settings picker yet.
- No graceful self-exit detection for WSL relays (cleanup relies on the kill
  path + Job Object backstop).

## Testing

See `docs/tests/windows-wsl-port-smoke.md`. Verified on Windows 11 + WSL2:
WSL runners + missions, native codex on the host, multi-window (Ctrl+N / menu /
sidebar, duplicate-subject overlay), CJK IME, platform shortcuts, and relay
cleanup on quit / force-kill.

> Branched off v0.2.10; upstream `main` has since moved, so this will likely
> need a rebase before merge.


yicheng47:
Review findings

Must fix:

1. `src-tauri/src/session/wsl/install.rs:32` embeds `../../../binaries/runner-agent-cli-linux-x86_64`, but the PR does not add a tracked file or a build/CI step that creates it. `git ls-files src-tauri/binaries` is empty, `test -e src-tauri/binaries/runner-agent-cli-linux-x86_64` fails, and the existing staging paths only build/stage the active target sidecars (`scripts/stage-runner-cli.mjs:66-101`, `.github/workflows/ci.yaml:63-70`). On a Windows build this is a compile-time `include_bytes!` failure before the best-effort installer can run. Please add a deterministic Linux runner-agent-cli staging/cross-build step for the Windows build path, or change this to a runtime-loaded artifact with a real packaged source.

2. `src/components/AppShell.tsx:62-73` now maps the sidebar toggle to `Ctrl+S` on Windows via `isModKey`, but it still explicitly lets xterm textarea focus through. Because the listener is registered on `window` capture, pressing `Ctrl+S` inside a terminal on Windows will be prevented and will toggle the sidebar instead of reaching the PTY as the shell control character. This contradicts the invariant documented in `src/lib/platform.ts:5-9` and the terminal handler, which only intercepts `Meta` shortcuts. Please preserve terminal Ctrl input here, e.g. guard `isXtermInput && !e.metaKey` before preventing default, while keeping Ctrl shortcuts for non-terminal app focus.

Checks run:
- `pnpm run lint` passed with warnings only: hook dependency warnings in `StartChatModal`, `MissionWorkspace`, `RunnerChat`, and fast-refresh warnings in `i18n.tsx`.
- `make test-rust` passed outside the sandbox. The first sandboxed run failed only because the MCP Unix-socket bind test hit `Operation not permitted`.
- GitHub reports no checks for this branch.

Review is not clean until the two must-fix items are addressed.

yicheng47:
Parking this rather than merging — it's a complete, working port, but it's branched from v0.2.10 and main has since reworked settings, sessions, and large parts of the UI (0.3.x line), so a rebase would effectively be a rewrite. Windows support is now tracked in #278 with this PR as the reference implementation; the `windows-port` branch stays as-is for that purpose.

Lycheen97:
Both must-fix items are addressed, plus the fallout they uncovered — pushed as six commits (9021fb2..96a6d3d).

**1. Embedded Linux CLI** (63dd41c): `stage-runner-cli.mjs` now cross-builds `runner-agent-cli` for `x86_64-unknown-linux-musl` whenever the app target is Windows and stages it at `binaries/runner-agent-cli-linux-x86_64`. musl + rustup's bundled `rust-lld` means no Linux toolchain is needed on the host; setup is a one-time `rustup target add x86_64-unknown-linux-musl`, and the script fails with that exact instruction when the target is missing. `src-tauri/build.rs` additionally guards the file with an actionable error instead of a bare `include_bytes!` failure. A new **Backend (Windows)** CI job stages the sidecars + embedded CLI on a clean `windows-latest` runner and runs `cargo clippy --workspace --all-targets -- -D warnings` — which also compiles all the `cfg(windows)` code the ubuntu job never sees. Verified locally by deleting the hand-staged binary and rebuilding from scratch; the staged musl ELF was executed inside WSL Ubuntu.

**2. Terminal Ctrl+S** (412f781): the sidebar listener now bails on xterm-focused events unless `metaKey` is held, so Ctrl+S / Ctrl+\ reach the PTY on Windows per the `platform.ts` invariant. ⌘S inside the terminal still toggles on macOS.

**Fallout — the Windows test suite had never actually run** (6edeaf6, 623d7af, ea4f8d9): the lib-test exe was dying at load with STATUS_ENTRYPOINT_NOT_FOUND — test binaries import comctl32 v6's `TaskDialogIndirect` but tauri_build embeds the common-controls manifest for bin targets only. So every cfg(windows) test, including this port's own `session::wsl` tests, had never executed anywhere. build.rs now feeds a shared `windows-app.manifest` (same content as tauri's default) to the linker for every executable. Actually running the suite surfaced:

- A real bug: `win_to_wsl` never reached its UNC branch — a slash-normalized `//wsl$/...` path starts with `/` and hit the POSIX passthrough first, so a project living in the distro filesystem got a broken cwd. UNC parsing now runs before the passthrough.
- `EventLog`'s tail repair and partial-write rollback failed on Windows: append-mode handles lack FILE_WRITE_DATA, so every `set_len` returned access-denied (the rollback silently, via `let _ =`). Writer handles now open in write mode and seek to EOF under the exclusive lock — equivalent behavior on unix.
- Unix assumptions in tests: separator-dependent substring assertions (cli_install), `Path::join` in `expand_home` producing `\` in bash-bound PATH strings, and the four PTY tests that spawn `/bin/sh` / `/bin/cat` (now `cfg(unix)`; the Windows spawn paths are covered by the smoke plan).
- One cross-platform flake: router tests counted `inject_and_submit`'s detached 80ms submit-`\r` in `len()` assertions, so results depended on scheduler timing. The recording injector's helpers now project the submit keystroke away.

Checks on Windows 11: `cargo test --workspace` green (runner lib 303/303, three consecutive runs), `cargo fmt --all --check` clean (the fmt sweep in ea4f8d9 is pure reformatting the check was failing on), `tsc --noEmit` passes, eslint unchanged (same 6 pre-existing warnings). The debug exe was launched as a runtime sanity check after the manifest change.

Clippy `-D warnings` is also clean on the Windows host now (96a6d3d): the deliberately-cross-platform MCP modules get a module-level `cfg_attr(windows, allow(dead_code))` (per their own "keep compiled so it can't rot" note) rather than cfg-gating the surface; the unix-only login-shell machinery in `shell_path` is `cfg(unix)`-gated; and the write-mode `OpenOptions` in runner-core got an explicit `.truncate(false)` — that last one would otherwise have tripped `suspicious_open_options` on the ubuntu gate as well. The Windows CI job runs clippy rather than plain check to keep it that way.

Issue #259, opened 2026-07-05
bug: duplicate-subject overlay misbehaves for multi-pane tabs on same-tab select


yicheng47:
## Description

The Arc-style duplicate-subject ("Open in another window") overlay behaves incorrectly when the selected chat lives in a multi-pane tab that is also open in another window.

- The overlay is computed and rendered **per pane** — `ChatPaneGroup.tsx` calls `secondaryFor(sid)` for each leaf and mounts a `DuplicateSubjectOverlay` inside that pane's body (`src/components/ChatPaneGroup.tsx:145-147`, `:251-260`). So a multi-pane tab that's duplicated in another window stacks **N separate overlay cards** (one per pane) instead of one tab-level notice.
- Because pane/tab layout is not synced across windows (see companion bug), the two windows **disagree about tab membership**. `isSecondaryFor` (`src/lib/windowFocus.ts:157-180`) then flips different panes to `secondary` in each window, so selecting the same tab another window drives gates individual panes read-only inconsistently and confusingly.

## Expected behavior

Selecting a chat/tab that another window already drives should present a single, coherent ownership state for the tab — one clear "another window owns this" affordance and consistent read-only gating — not a per-pane pile of overlay cards, and not a scoping that depends on each window's private idea of what's in the tab.

## Steps to reproduce

1. Open two windows (⌘N).
2. In window A, open a multi-pane tab (e.g. chats X and Y side by side).
3. In window B, select/open the same tab (or one of its member chats), then focus window B so it becomes primary.
4. Observe window A: each pane flips to its own "Open in another window" card (stacked overlays), and which panes are gated read-only is inconsistent with what window B actually shows.

## Relevant code

- `src/components/DuplicateSubjectOverlay.tsx` — the overlay card.
- `src/components/ChatPaneGroup.tsx:141-260` — per-pane `secondaryFor` + per-pane overlay mount (`:145-147`, `:251-260`); terminal gate `renderTerminalPane` (`:311-312`).
- `src/lib/windowFocus.ts:157-180` — `isSecondaryFor` secondary detection.
- `src/pages/RunnerChat.tsx:318-360` — per-pane subject reporting + `secondaryBySession` map; `reportSubjectsNow` on pane fill (`:1827`).

Downstream of the cross-window layout-sync bug; also needs its own fix to scope the overlay at the tab level (or dedupe to a single card) rather than one card per pane.

## Environment
- Version: 0.3.3

yicheng47:
[...]

PR #260, opened 2026-07-05
fix(ui): sync multi-pane tab layout across windows

### LINKED ISSUE #258: bug: multi-pane tab state not synced across windows
## Description

The pane/tab layout state (multi-pane tabs from the tab accordion, impl 0023) does not refresh to other open windows. When you change tab/pane state in one window — split a chat into panes, group/regroup tabs, switch the active tab, collapse a tab — the other open windows keep rendering their stale tab grouping.

Root cause: the layout store is **per-window and in-memory** with no cross-window propagation.

- `src/lib/paneLayout.ts` holds `layouts`/`activeIndex` as module-level vars, mutated through `setLayoutSet` (~`src/lib/paneLayout.ts:557`), which only notifies **in-window** React subscribers via `useSyncExternalStore`.
- Its only persistence path is localStorage key `runner.chat.layout`, and that is **main-window-only** (the `persistToStorage` gate, `src/lib/paneLayout.ts:437-443`) and write-only — no window registers a `storage` listener for that key, and secondary windows never write it.
- There is no Tauri `emit`/`listen`, no `BroadcastChannel`, and no DB round-trip for layout.

Contrast with pins, which DO sync across windows: `session_pin` writes the DB then `app.emit("session/updated", …)` (`src-tauri/src/commands/session.rs:447`), and every window's Sidebar listens and re-fetches (`src/components/Sidebar.tsx:569-576`). paneLayout has no equivalent — a direct consequence of impl 0023 keeping tabs frontend-only; nothing wired the cross-window channel.

## Expected behavior

A tab/pane layout change in one window converges to every other window showing the same surface, the way pin/rename/archive already do — so all windows agree on tab membership, the active tab, and collapsed state.

## Steps to reproduce

1. Open two Runner windows (⌘N).
2. In window A, split a chat into a multi-pane tab (or group two chats into one tab).
3. Observe window B still shows the old, ungrouped layout — the tab state never refreshes there.

## Relevant code

- `src/lib/paneLayout.ts` — per-window store: `setLayoutSet` (~:557), `persistToStorage` main-window gate (:437-443), localStorage key `runner.chat.layout` (:320), `persistLayoutSet`/`readPersistedLayoutSet` (:508/:495).
- `src/components/Sidebar.tsx:284-285` — `usePaneLayout`/`usePaneLayouts` (in-window `useSyncExternalStore` only, no cross-window listener).
- Reference sync path to mirror: `src-tauri/src/commands/session.rs:447` (`app.emit("session/updated")`) + `src/components/Sidebar.tsx:569-576` (listen → refresh).

Fix direction: broadcast a layout-changed Tauri event on every `setLayoutSet` mutation and reload it in each window's `paneLayout.ts` (or persist to the DB + refresh-on-broadcast like pins). Coupled to the duplicate-subject overlay bug (windows disagreeing on tab membership).

## Environment
- Version: 0.3.3

yicheng47:
## Summary

Fixes #258 — a multi-pane tab/pane layout change made in one window now converges to every other open window, and a newly-opened window shows the existing structure, the way pins already do.

The pane/tab layout store (`src/lib/paneLayout.ts`) was **per-window and in-memory**: `setLayoutSet` only notified in-window `useSyncExternalStore` subscribers, and its only persistence was the **main-window-only, write-only** localStorage key `runner.chat.layout`. No Tauri emit/listen, no BroadcastChannel, no DB round-trip — so split/group/regroup/collapse in window A never reached window B, and a freshly-opened window started blank.

## Approach

Mirror the pins convergence path, kept **frontend-only** since tabs are frontend-only (impl 0023):

- `setLayoutSet` splits into `applyLayoutSet` (normalize → swap → persist → notify) and a thin `setLayoutSet` that also calls `broadcastLayoutSet()`.
- `broadcastLayoutSet` emits the serialized `PersistedLayoutSet` over a `chat/layout-changed` Tauri event, tagged with the origin window label.
- Every window listens, **ignores its own echo** (Tauri `emit` delivers to the sender too) by comparing `origin` to its own label, then hydrates via `hydrateLayoutSet` → `applyLayoutSet`, which **never re-broadcasts** → no feedback loop.
- **Cold start**: localStorage is shared across webview windows, so reads are open to every window (writes stay main-only). A newly-opened secondary window loads the main window's persisted structure immediately, then tracks live changes over the broadcast channel. Its own edits still reach storage via broadcast → main hydrate → main persist, so there is no clobbering.
- **Off-Tauri (dev browser)**: the window label resolves once in a try/catch (null); the emit and the module-load listener both no-op.

### What converges vs. what stays per-window

Tab **membership, pane→session assignments, split sizes, collapsed state, and group names** converge. The **active tab (`activeIndex`) stays per-window**: it is view state bound to each window's route, and local mutators (`closePane`/`focusPane`/`setGroupName`) target `currentLayout()`. On hydrate, a window keeps its own active tab by re-finding, in the incoming set, the tab holding **its route session** — recorded via a dedicated `setRouteAnchorSession` called only from RunnerChat's URL effect (so a sidebar pick or background-tab rename can't move the anchor). Anchoring on the route session rather than a tab's `focusedPaneId` matters because focus itself is synced.

Two supporting fixes surfaced in review:
- Hydrated members are marked **fresh** so RunnerChat's vanished-session sweep doesn't evict a chat just created in another window before this window's row refresh sees it (which would rebroadcast the stale layout).
- `recordSplitSizes` now **broadcasts** on drag-end so sizes converge and a secondary-window resize reaches the main window's localStorage.

## Scope

Fixes **only** #258. Does not touch #259 (duplicate-subject / conflict overlay). No Rust changes.

## Checks

- `pnpm exec tsc --noEmit` — clean
- `pnpm run lint` — clean
- `pnpm exec vitest run` — 70 passed (9 sync cases: membership convergence, route-anchor preservation, anchor-vs-synced-focus, background-tab anchor isolation, follow-anchor-on-move, fresh-marking, collapsed convergence, malformed no-op, secondary-window cold start)

Reviewed on the working-tree diff across rounds before/while this PR was open.

### DIFF
--- a/src/lib/paneLayout.test.ts
+++ b/src/lib/paneLayout.test.ts
@@ -21,6 +21,7 @@ import {
   resetPaneLayoutsForTest,
   serializeLayout,
   serializeLayoutSet,
+  setRouteAnchorSession,
   setSizesPure,
   setTabCollapsed,
   subscribePaneLayout,
@@ -307,14 +308,16 @@ describe("setTabCollapsed", () => {
 
 describe("hydrateLayoutSet (cross-window sync)", () => {
   afterEach(() => {
+    vi.useRealTimers();
     resetPaneLayoutsForTest();
   });
 
-  it("replaces the store with a set broadcast from another window and notifies subscribers", () => {
-    // Local state: a single tab on chat A.
-    applyPreset("single", "A", ["A"]);
+  it("converges tab membership from another window and notifies subscribers", () => {
+    // Local state: one tab {A,B}, focused on A (this window's active tab).
+    applyPreset("cols-2", "A", ["A", "B"]);
 
-    // A remote window's serialized set: two tabs, second one active.
+    // A remote window's serialized set: our tab plus a second one it made
+    // active.
     const remote = serializeLayoutSet(
       [
         applyPresetPure("cols-2", "A", ["A", "B"]),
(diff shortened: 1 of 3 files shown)


yicheng47:
Closing until the working-tree diff review is complete — per an updated crew workflow, reviewer sign-off precedes the PR. Will reopen this same branch once review is clean and re-run CI.

PR #427, opened 2026-08-21
feat(terminal): keep session terminals alive

yicheng47:
## Summary

- keep one long-lived alacritty Term per live backend session and feed it synchronously from raw SessionEvents output
- remove the backend output replay ring, snapshots, synthetic prefix, resize purge, and rows-nudge coherence machinery
- make chat and mission panes borrow registry terminals across tab and route changes; replace on spawn and release on exit/archive
- persist only the final terminal size once per resize storm
- queue human delivery across a stopped/resuming PTY, warn once in the feed, and flush on respawn

## Verification

- make verify
- reviewer independently reran make verify: 542 backend, 123 app, 50 app-lib, 27 terminal tests plus integrations; clippy and fmt clean
- nightly daily-drive smoke passed
- real PTY 20-cycle start/kill test ends with 0 live terminals after every cycle

## Visible lifecycle note

The Term is intentionally released on session exit per M6.8. A stopped chat or mission pane therefore shows an empty terminal background behind the Ended/Resume card; resume creates a fresh Term for the fresh child.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5052,12 +5052,12 @@ dependencies = [
  "alacritty_terminal",
  "anyhow",
  "base64",
+ "log",
  "regex",
  "runner-backend",
  "serde",
  "serde_json",
  "tempfile",
- "tokio",
 ]
 
 [[package]]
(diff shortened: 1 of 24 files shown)


yicheng47:
Closing per maintainer request; the reviewed change will be merged directly into gpui-nightly.

